PR #3, opened 2019-07-08
Add security groups attribute, into example for creating OS vm.

Jst10:


examples/service.yaml
```diff
@@ -16,6 +16,9 @@ node_types:
       network:
         type: string
         description: OpenStack network id (network names are not accepted)
+      security_groups:
```
tadeboro: Is there any reason for not using a `list` type here? Using a comma-separated string here feels like a hack.

I would suggest writing this down as:

    security_groups:
      description: OpenStack ...
      type: list
      entry_schema:
        type: string

.gitignore
```diff
@@ -8,3 +8,4 @@ eggs/
 dist/
 
 venv/
+.idea/
```
tadeboro: Since this is not opera-specific ignore, it would make more sense to add it into `$XDG_CONFIG/git/ignore` where all user-specific ignores should be (like editor backup buffers, config folders, etc.). As a bonus, you only need to add it once and it will be ignored in all of your git repos.

examples/service.yaml
```diff
@@ -87,6 +91,7 @@ topology_template:
         image: 9ea4856a-32b2-4553-b408-cfa4cb1bb40b
         flavor: d3046a41-245a-4042-862e-59568e81f8fa
         network: 753940e0-c2a7-4c9d-992e-4d5bd71f85aa
+        security_groups: default
```
tadeboro: If we decide to make `security_groups` a list of strings, this will need to be adjusted accordingly.

examples/service.yaml
Jst10: os_server module supports both `list` and `comma-separated string`.  But when I tried  to use `list` type,  xOpera throwed me an exception:
```
Deploying instance model ...
  Processing vm.0 ...
    Executing primary: playbooks/vm/create.yml ...
Traceback (most recent call last):
  File "/home/jst10/opera/.venv/bin/opera", line 11, in <module>
    sys.exit(main())
  File "/home/jst10/opera/.venv/lib/python3.6/site-packages/opera/cli.py", line 41, in main
    return args.func(args)
  File "/home/jst10/opera/.venv/lib/python3.6/site-packages/opera/commands/deploy.py", line 37, in deploy
    instances.deploy()
  File "/home/jst10/opera/.venv/lib/python3.6/site-packages/opera/instances.py", line 142, in deploy
    self.nodes[id].deploy()
  File "/home/jst10/opera/.venv/lib/python3.6/site-packages/opera/instances.py", line 61, in deploy
    start=("starting", "started"),
  File "/home/jst10/opera/.venv/lib/python3.6/site-packages/opera/instances.py", line 51, in execute_workflow
    success, attributes = operation.run()
  File "/home/jst10/opera/.venv/lib/python3.6/site-packages/opera/operations.py", line 18, in run
    k: v.eval(self.instance) for k, v in self.inputs.items()
  File "/home/jst10/opera/.venv/lib/python3.6/site-packages/opera/operations.py", line 18, in <dictcomp>
    k: v.eval(self.instance) for k, v in self.inputs.items()
  File "/home/jst10/opera/.venv/lib/python3.6/site-packages/opera/types.py", line 188, in eval
    return getattr(reference, function)(*arguments)
  File "/home/jst10/opera/.venv/lib/python3.6/site-packages/opera/instances.py", line 72, in get_property
    return self.template.get_property(*path)
  File "/home/jst10/opera/.venv/lib/python3.6/site-packages/opera/types.py", line 521, in get_property
    return prop.eval(self)
AttributeError: 'Pass' object has no attribute 'eval'

```
Because of that I assumed that `list` type is not supported yet in xOpera, so this is reason for using   `comma-separated string`.

.gitignore
Jst10: I never considered to use global git configuration, but in this case seem to be the best choice, thank you.

examples/service.yaml
Jst10: After looking into code, I think that this:
```
    def eval(self, _reference):
        return self.data
```
is missing from `Pass` class in `opera/types.py:76`.

tadeboro: The `Pass` class is a marker for **this is not implemented yet, but I would still like to parse at least something**. If we make it play too nicely with the rest of the code, I might lose interest in fixing bugs and adding new functionality ;)

I will fill the bug on the opera repo and merge this as it is.

PR #6, opened 2019-08-05
Add YAML 1.2 loader

tadeboro:
To correctly parse the TOSCA documents, we created a YAML 1.2 compatible loader. And while we were at it, we also added location information to every piece of data that loader produces. Location information allows us to write more informative error messages when reporting issues in TOSCA documents.


tests/opera/parser/yaml/test_constructor.py
```diff
@@ -0,0 +1,140 @@
...
+    @pytest.mark.parametrize(
+        "value", [
+            "abc", "1.2.3", "NaN", "INF", ".NAn", " ", "   ", "\n", "\t",
+            "1 2", "https://url", ":bare", "multi\n  line",
+        ],
+    )
+    def test_construct_str(self, value):
+        mark = Mark(None, None, 0, 0, None, None)
+        node = ScalarNode(None, value, start_mark=mark)
+        res = Constructor("float").construct_yaml_str(node)
```
matejart: Should this be `str` instead of `float`? Just to keep the appearances.

tests/opera/parser/yaml/test_constructor.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def test_construct_seq(self):
+        mark = Mark(None, None, 4, 3, None, None)
+        children = [
+            ScalarNode("tag:yaml.org,2002:int", str(i), start_mark=mark)
+            for i in range(4)
+        ]
+        node = SequenceNode(None, children, start_mark=mark)
+        # Sequences are single-element generators (lazy parsing) that we need
+        # to exhaust. Simplest way of doing this is to assing to a single
```
matejart: I'd google the word "assing" to see if it means what it means, but it sounds NSFW.

src/opera/parser/yaml/resolver.py
```diff
@@ -0,0 +1,62 @@
...
+        pass
+
+    def ascend_resolver(_self):
+        pass
+
+
+# From https://yaml.org/spec/1.2/spec.html#id2804923
+Resolver.add_implicit_resolver(
+    "tag:yaml.org,2002:null",
+    re.compile(r"^(~|null|Null|NULL|)$"),
```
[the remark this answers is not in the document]
tadeboro: Whitespace handling, according to the YAML standard, is performed by the composer. The only way we can get the trailing space here is we quote the `null` in the YAML document, but in this case, we should tag it as a string anyway.

PR #8, opened 2019-08-13
Import resolution process

tadeboro:
This batch of commits brings parser far enough that we can resolve the imports in the TOSCA documents.

And since now we can parse and check the syntactic validity of TOSCA documents, we quickly added a `validate` command to the `opera` CLI interface.

NOTE: This is marked as WIP because we need to rebase it on master once #7 is merged.


src/opera/stdlib/v_1_3.yaml
```diff
@@ -0,0 +1,668 @@
...
+      algorithm:
+        type: string
+        required: false
+        status: experimental
+    capabilities:
+      client:
+        type: tosca.capabilities.Endpoint.Public
+        occurrences: [0, UNBOUNDED]
+        description: >
+          The Floating (IP) client’s on the public network can connect to.
```
matejart: clients

tadeboro: Just for reference: this is copy-pasted from TOSCA standard.

PR #13, opened 2019-09-27
Continue refactor: linting, CSARs, parsing.

sstanovnik:
What a big one. Feel free to close and pick cherries from this tree; this is the product of slowly exploring the codebase.

There are a lot of notable changes:

* `python -m opera.cli <args>` now works,
* `mypy` and type hints,
* a `pylint` configuration (120 line length which might be an issue for you based on existing code),
* `make lint` that runs both of the above on src and tests,
* console-friendly logging,
* an overhaul of the parsing structure, including support for ZIP-based and directory-based CSARs; described below.

Logging is done through Python's `logging`. A `LOG_LEVEL` environment variable controls the output, which is still console-friendly. When `INFO`, only stdout is used and _regular_ console messages are printed, i.e. without module names and timestamps. But when `DEBUG`, a separate stream to stderr is used to write all messages (DEBUG and above) with module names and timestamps. Should turn into a `--verbose` switch in the future.

The TOSCA standard is _a bit_ hard to read so things I've done may be incompatible, especially with respect to path references. I've assumed that relative paths are always relative to the CSAR root and that, if there is an import with an absolute path, it can only have absolute paths in its own imports. Relative path access is otherwise constrained inside a CSAR.

If you look at 1e78a5d0a025e0390492c1723d1bc24c2a8b925d, it introduces three CSAR types, which are a bit hackish:

* `ZipToscaCsar`,
* `DirectoryToscaCsar`,
* `VirtualToscaCsar`.

Zip is the zip-based format from the standard. Files are streamed directly from it, I didn't extract it. Because I only ran `opera validate` and the tests, I don't know what implications not extracting has on using Ansible to actually do stuff.

Directory is the xopera extension where the CSAR, still conforming to the standard, except for the metadata directory, is extracted.

Virtual is where a single file is targeted; the CSAR structure is disregarded and the "CSAR root" is its containing folder.

What I wanted to do here is consolidate filesystem access and access control through ToscaCsar to make passing around a `base_path` a bit cleaner. `ToscaCsar.load(path)` creates one of these depending on whether the path is a zip file, a directory, or a bare file (assumed to be a service template).


src/opera/commands/deploy.py
```diff
@@ -19,29 +19,30 @@ def add_parser(subparsers):
...
     status = 0
 
-    print("Loading service template ...")
+    logger.info("Loading service template ...")
     try:
-        ast = tosca.load(Path.cwd(), PurePath(args.csar.name))
-        print("=======================")
-        tmpl = template.build(ast)
-        print(tmpl)
+        # ast = tosca.load(Path.cwd(), PurePath(args.csar.name))
```
matejart: Is the deploy now fully disabled? Or have I missed something and it's started from somewhere else?

matejart:
TIL the parmeter and method return value declarations.

matejart:
Thank you, @sstanovnik, for your work!

Here are a few comments:

* Streaming the zip contents is fine for validation.
* For other actions, the package needs to be unpacked and all its contents present in a regular file system. The artifacts _will_ reference and import eachother, so they need to be present at all times during each workflow.
* The previous bullet opens a question: do we unpack the csar for each invocation of a workflow, then throw it away? Or should there be some kind of persistence between calls? Of course the most trivial and useful strategy at this point is to treat them as temporary. The user should be responsible for optimizing frequent call performance.
* Absolute paths: we discussed this before and decided that the best practice is to ban them. Everything should be in the CSAR, and any attempts, benign or malicious, to access filesystem outside of that should be blocked.

I've only had a cursory look at the code, leaving comments where I thought would be necessary.

src/opera/csar.py
```diff
@@ -1,13 +1,207 @@
...
+        """
+        logger.debug("Validating CSAR.")
+        files = self.members()
+
+        relative_root = pathlib.PurePath(".")
+        if any(f for f in files if
+               f.parent == relative_root
+               and f.name == ToscaCsar.METADATA_DIRNAME
+               and self._member_is_dir(f)):
+            raise UnsupportedToscaFeatureError("xopera does not support CSAR structures with a metadata directory.")
```
matejart: `xOpera`

matejart: This applies to all text occurences in comments, logger outputs and error messages

tests/opera/test_csar.py
```diff
@@ -0,0 +1,78 @@
...
+    def test_load_missing_metadata(self, mode, tmp_path: pathlib.Path):
+        with pytest.raises(CsarValidationError):
+            _base_loader(_RESOURCE_DIRECTORY / "csar/missingmetadata/", mode, tmp_path)
+
+    @pytest.mark.parametrize("mode", [CsarTestMode.ZIP, CsarTestMode.DIRECTORY])
+    def test_load_multiple_root_files(self, mode, tmp_path: pathlib.Path):
+        with pytest.raises(CsarValidationError):
+            _base_loader(_RESOURCE_DIRECTORY / "csar/multipleroot/", mode, tmp_path)
+
+    @pytest.mark.parametrize("where", ["../../../../../../../../../../../../../../../../etc/hosts", "/etc/hosts"])
```
matejart: Could you also include `"../../etc/hosts"` which should also already fall out of the csar root path?

src/opera/parser/parser.py
```diff
@@ -0,0 +1,159 @@
...
+        return tosca_document_parsed
+
+    def _load_stdlib(self, tosca_version: str) -> TDocument:
+        stdlib_yaml = stdlib.load(tosca_version)
+        return self.document_type.parse(stdlib_yaml)
+
+    def _merge_imports(self, csar: ToscaCsar, document: TDocument):
+        """
+        Import paths can be absolute or relative.
+          - absolute paths can be anywhere on the disk
```
matejart: With CSAR support, we shouldn't do this anymore. Everything should be on the disk and within the CSAR root directory. Nothing outside it should exist for the merge or the artifacts.

tests/opera/parser/tosca/test_path.py
```diff
@@ -1,106 +1,9 @@
...
-        symlink = tmp_path / "symlink"
-        symlink.symlink_to(file)
-
-        with pytest.raises(ParseError):
-            Path(pathlib.PurePath("symlink"), None).resolve_path(tmp_path)
-
-    @pytest.mark.parametrize("path", [
-        "..", "../a", "../../a/b", "a/../..", "a/b/../c/../../..",
-        "/..", "/../a", "/../../a/b", "/a/../..", "/a/b/../c/../../..",
-    ])
```
matejart: These cases are important to test in the unit tests. Please make sure that the equivalent of them exists somewhere.

src/opera/commands/deploy.py
sstanovnik: I haven't touched the deploy and undeploy code paths, apart from commenting out code so the parser and linter don't complain. The refactor before mine was only partial, and left the two in a non-working state; see 5cad99bbcea939df40d0b0cbedef043fac5e179c. Perhaps @tadeboro has some plans for this.

tests/opera/test_csar.py
sstanovnik: I've added tests for non-existent files outside the CSAR, relatively closer in the tree, via relative and absolute paths, which I think you were aiming for.

src/opera/parser/parser.py
sstanovnik: Great. Do you want me to fix this in this PR or make a new one? It's quite a significant change which would benefit from a separate review process.

This is another divergence from the TOSCA standard, right? There are instances of absolute paths e.g. in [Simple Profile 1.3](https://docs.oasis-open.org/tosca/TOSCA-Simple-Profile-YAML/v1.3/TOSCA-Simple-Profile-YAML-v1.3.html) in Section 4.8.1.3.1.

tests/opera/test_csar.py
matejart: Actually, what I was aiming at is the files that _do_ exist outside the tree root or inside the peer trees.

src/opera/parser/parser.py
matejart: Ugh, so I can craft a CSAR that imports something off the file system that's not within the CSAR? I don't know, it feels like there should be a switch for this kind of icky compliance.

In any case, I'd make it a new PR, but let's discuss it with @tadeboro first.

tests/opera/test_csar.py
sstanovnik: What do you mean? `../../etc/hosts` would not exist in the scope of this test, as it would point to `<resource-directory>/csar/barebones/../../etc/hosts -> <resource-directory>/etc/hosts`. I've used a (hopefully always valid) relative path and an absolute path to a file that, again hopefully, always exists. If the two hopefullies are what is bothering you, I've now explicitly created a temporary file that must exist and written separate tests for absolute and relative access.

tests/opera/parser/tosca/test_path.py
sstanovnik: Will the out of bounds access control tests suffice? Otherwise we're only testing whether `pathlib` and the underlying filesystem work, and we're kind of supposed to trust that.

sstanovnik:
Re: streaming the zip file instead of extracting. Do you want to continue this in a separate PR? If a directory or file (the `VirtualToscaCsar` option) is targeted with xOpera, things will still work, as this PR doesn't break anything in that regard. The only functional thing it adds is zipped CSAR validation, but CSARs weren't even supported before.

As for solutions, there are two:
1) unzipping the CSAR into a temporary directory every time
2) use `avfs` or similar to make a read-only mount of the zip

Not unarchiving every time will not work, as we would have to verify or trust that the user hasn't changed anything in the extracted files, at which point it just becomes easier to overwrite everything.

The second option is very fancy, but includes a rather heavy system dependency. Also a possible maintenance nightmare.

Therefore, for simplicity's sake, I'd drop the streamed zip support and only have the CSAR be used expanded on disk, but still with the option to target either a zip, a yml or a directory with the CLI. All this in another PR though, same as the absolute path prohibition.

PR #20, opened 2020-02-06
Add Acknowledgement

matejart:
We thank RADON project for having partially funded this work.


cankarm:
I like that, thanks @matejart for adding this! 

Maybe I would suggest to add grant number of SODALITE in addition to RADON. The URL links to the projects are optional.

matejart:
Thank you for approvals. For now we keep RADON only.

PR #23, opened 2020-02-12
Add outputs command

### LINKED ISSUE #22: Support for outputs in topology_template
Current version of xOpera (0.5.2) does not support output section in the topology template. The following example is accepted by the application, but nothing happens. 

Example:
```
topology_template:
  node_templates:
    db_server:
      type: tosca.nodes.Compute
      capabilities:
        # Host container properties
        host:
          properties:
            # Compute properties
            num_cpus: 5
            mem_size: 2048  MB
            disk_size: 10 GB
            mem_size: 4096 MB
  outputs:
    disk_size:
      description: The size of the disk
      value: { get_attribute: [ db_server, disk_size ] }
```

Preferred/expected behavior:
```
$ opera outputs --yaml
disk_size:
  description:  The size of the disk
  value: 10 GB
```

tadeboro:
We can use newly added command for retrieving service template outputs in a machine-readable way.

Fixes #22

### DIFF
--- a/src/opera/template/node.py
+++ b/src/opera/template/node.py
@@ -103,5 +103,19 @@ class Node:
             ))
         return requirements[0].target.get_property(["SELF"] + rest)
 
+    def get_attribute(self, params):
+        host, prop, *rest = params
+
+        if host != "SELF":
+            raise DataError(
+                "Accessing non-local stuff is bad. Fix your service template."
+            )
+        if host == "HOST":
+            raise DataError("HOST is not yet supported in opera.")
+        if len(self.instances) != 1:
+            raise DataError("Cannot get an attribute from multiple instances")
+
+        return next(iter(self.instances.values())).get_attribute(params)
+
     def get_input(self, params):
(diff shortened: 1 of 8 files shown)


src/opera/template/node.py
```diff
@@ -103,5 +103,19 @@ def get_property(self, params):
...
 
+    def get_attribute(self, params):
+        host, prop, *rest = params
+
+        if host != "SELF":
+            raise DataError(
+                "Accessing non-local stuff is bad. Fix your service template."
+            )
+        if host == "HOST":
+            raise DataError("HOST is not yet supported in opera.")
```
matejart: Is this just a placeholder for the dev to remember about the HOST? Because from how I see it, this condition is never evaluated as `True`.

tadeboro: That is the exact purpose of this code: to be able to grep for `HOST` and get back all of the relevant places that might need updating when/if we add support.

PR #24, opened 2020-02-14
Make ouputs optional.

anzoman:
In the current version (0.5.3) of xOpera, outputs are mandatory. I would suggest a quick fix to make them optional. @tadeboro could you take a look?


anzoman:
@tadeboro I added a bit more descriptive commit message based on your suggestion.

PR #29, opened 2020-04-03
Requirement definition fixes

anzoman:
Resolve bug that was found in TOSCA parser for service template and add more meaningful error messages for relationship requirements.

### DIFF
--- a/src/opera/parser/tosca/v_1_3/requirement_assignment.py
+++ b/src/opera/parser/tosca/v_1_3/requirement_assignment.py
@@ -25,3 +25,9 @@ class RequirementAssignment(Entity):
         if isinstance(yaml_node.value, str):
             return Node({Node("node"): yaml_node})
         return yaml_node
+
+    @classmethod
+    def validate(cls, yaml_node):
+        super().validate(yaml_node)
+        if "relationship" in yaml_node.bare and not isinstance(yaml_node.bare["relationship"], str):
+            cls.abort("Expected a relationship template name as a 'relationship' value.", yaml_node.loc)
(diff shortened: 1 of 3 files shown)


src/opera/parser/tosca/v_1_3/requirement_definition.py
```diff
@@ -22,3 +22,9 @@ def normalize(cls, yaml_node):
         if isinstance(yaml_node.value, str):
             return Node({Node("capability"): yaml_node})
         return yaml_node
+
+    @classmethod
+    def validate(cls, yaml_node):
+        super().validate(yaml_node)
+        if "relationship" in yaml_node.bare and not isinstance(yaml_node.bare["relationship"], str):
+            cls.abort("Expected a relationship string reference.", yaml_node.loc)
```
tadeboro, line 30: Maybe we can be more specific here: "Expected a relationship type name as a 'relationship' value."

anzoman: Ok, I chaged that as you suggested.

anzoman:
I also added the relationship validation to the requirement assignment so there will be no need to do this in a separate PR.

src/opera/parser/tosca/v_1_3/requirement_assignment.py
```diff
@@ -25,3 +25,9 @@ def normalize(cls, yaml_node):
         if isinstance(yaml_node.value, str):
             return Node({Node("node"): yaml_node})
         return yaml_node
+
+    @classmethod
+    def validate(cls, yaml_node):
+        super().validate(yaml_node)
+        if "relationship" in yaml_node.bare and not isinstance(yaml_node.bare["relationship"], str):
+            cls.abort("Expected a relationship type name as a 'relationship' value.", yaml_node.loc)
```
tadeboro, line 33: Here, we are expecting a relationship template name, not type.

anzoman: Expecting type was changed to expecting template.

PR #31, opened 2020-04-17
Remove the root __init__.py to enable PEP420.

sstanovnik:
[PEP420](https://www.python.org/dev/peps/pep-0420/) are Implicit Namespace Packages. Without breaking anything (assuming Python >= 3.3), it allows the `opera` namespace (previously a package) to be extended in other source trees.

Example.

* Create a new project that depends on xopera-opera - `pip install opera==0.5.5`.
* `mkdir -p opera/extension/`
* `touch opera/extension/__init__.py`
* `echo "import opera.commands" > opera/extension/thing.py`
* `python3 -m opera.extension.thing`
* Result
  * without PEP420 in this repository: `Error while finding module specification for 'opera.extension.thing' (ModuleNotFoundError: No module named 'opera.extension')`
  * with PEP420 in this repository: EOF

To be specific about my needs, I would like to `import opera` in the xOpera API package we're developing and _also_ use `opera.api` as a package to contain the API implementation.

I've tested `pip install -e . && opera` to work with the file missing.

cc @cankarm

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -25,7 +25,7 @@ classifiers =
 [options]
 package_dir =
   = src
-packages = find:
+packages = find_namespace:
 zip_safe = True
 include_package_data = True
 setup_requires =


anzoman:
Thanks for the PR @sstanovnik I understand that importing opera commands could be an important thing. Although your changes resolve ModuleNotFoundError, they might bring some damage. When you remove `__init.py__` from the src directory this seems to break the package. I tried to rebuild opera manually like this:

```
python3 setup.py sdist bdist_wheel
pip install dist/opera-0.5.6.dev4+ge4506ba-py3-none-any.whl (or pip install dist/opera-0.5.6.dev4+ge4506ba.tar.gz)
```

And when trying to run `opera˙ command from the console I get this import error:

error

Can you confirm that you get the same result when building the pip package manually?

tadeboro:
`setup.cfg` needs fixing too. Instead of using `packages = find:`, we need to use `packages = find_namespace:`.

tadeboro:
And a bit higher-level question: is making the `opera` package a namespace package necessary? The primary reason for using namespace packages is to allow the package content to come from more than one location on a disk. But the mentioned project does not seem to extend the `opera` package. It is just a regular application that uses the `opera` package.

sstanovnik:
Using `find_namespace:` did the trick, so now

```
pip uninstall -y opera \
    && rm -rf build/ dist/ \
    && python3 setup.py sdist bdist_wheel \
    && pip install dist/*.whl \
    && opera
```

works as expected.

@tadeboro how do you mean it doesn't use the `opera` package? The second line of my example does `mkdir -p opera/extension/`. It doesn't `touch opera/__init__.py` because it also can't include it to be able to use PEP420. Well, to be pedantic, it really doesn't use the opera _package_, it uses it as a _namespace_, if that's what you mean.

tadeboro:
> @tadeboro how do you mean it doesn't use the `opera` package? The second line of my example does `mkdir -p opera/extension/`. It doesn't `touch opera/__init__.py` because it also can't include it to be able to use PEP420. Well, to be pedantic, it really doesn't use the opera _package_, it uses it as a _namespace_, if that's what you mean.

I was looking at the xopera-api when I was talking about the project utilizing the namespace support. That package does not seem to add new functionalities to the opera package. Or I am missing some other project that is trying to add new pieces of functionality into the opera package?

PR #33, opened 2020-04-29
Add duplicate imports management

anzoman:
Currently opera throws duplicate key error when importing one template multiple times. This can be fixed by remembering which files have been already imported into the main TOSCA service template.

### DIFF
--- a/src/opera/parser/tosca/v_1_3/service_template.py
+++ b/src/opera/parser/tosca/v_1_3/service_template.py
@@ -56,24 +56,36 @@ class ServiceTemplate(Entity):
         }, yaml_node.loc)
 
     @classmethod
-    def parse(cls, yaml_node, base_path, csar_path):
+    def parse(cls, yaml_node, base_path, template_path, parsed_templates):
         service = super().parse(yaml_node)
-        service.visit("prefix_path", csar_path)
-        service.merge_imports(base_path)
-        return service
+        service.visit("prefix_path", template_path.parent)
+        parsed = service.merge_imports(
+            base_path, parsed_templates | set([template_path]),
+        )
+        return service, parsed
 
-    def merge_imports(self, base_path):
+    def merge_imports(self, base_path, parsed_templates):
+        parsed = parsed_templates.copy()
+
         for import_def in self.data.get("imports", []):
             import_def.file.resolve_path(base_path)
             import_path = import_def.file.data
+
+            if import_path in parsed:
+                continue
+
             with (base_path / import_path).open() as fd:
                 yaml_data = yaml.load(fd, str(import_path))
-            self.merge(ServiceTemplate.parse(
-                yaml_data, base_path, import_path.parent,
-            ))
+            ast, parsed = ServiceTemplate.parse(
+                yaml_data, base_path, import_path, parsed,
+            )
+            self.merge(ast)
+
         # We do not need imports anymore, since they are preprocessor
         # construct and would only clutter the AST.
         self.data.pop("imports", None)
+
+        return parsed
 
     def merge(self, other):
         if self.tosca_definitions_version != other.tosca_definitions_version:
--- a/tests/opera/parser/test_tosca.py
+++ b/tests/opera/parser/test_tosca.py
@@ -101,3 +101,42 @@ class TestLoad:
 
         doc = tosca.load(tmp_path, name)
         assert doc.data_types["my_type"]
+
+    def test_duplicate_import(self, tmp_path, yaml_text):
+        name = pathlib.PurePath("template.yaml")
+        (tmp_path / name).write_text(yaml_text(
+            """
+            tosca_definitions_version: tosca_simple_yaml_1_3
+            imports: [ template.yaml ]
+            """
+        ))
+        tosca.load(tmp_path, name)
+
+    def test_imports_from_multiple_levels(self, tmp_path, yaml_text):
+        name = pathlib.PurePath("template.yaml")
+        (tmp_path / name).write_text(yaml_text(
+            """
+            tosca_definitions_version: tosca_simple_yaml_1_3
+            imports:
+              - subfolder/a.yaml
+              - subfolder/b.yaml
+            """
+        ))
+        (tmp_path / "subfolder").mkdir()
+        (tmp_path / "subfolder/a.yaml").write_text(yaml_text(
+            """
+            tosca_definitions_version: tosca_simple_yaml_1_3
+            imports:
+              - b.yaml
+            """
+        ))
+        (tmp_path / "subfolder/b.yaml").write_text(yaml_text(
+            """
+            tosca_definitions_version: tosca_simple_yaml_1_3
+            data_types:
+              my_type:
+                derived_from: tosca.datatypes.xml
+            """
+        ))
+
+        tosca.load(tmp_path, name)
--- a/tests/opera/parser/tosca/v_1_3/test_service_template.py
+++ b/tests/opera/parser/tosca/v_1_3/test_service_template.py
@@ -21,7 +21,7 @@ class TestNormalizeDefinition:
 
 
 class TestParse:
-    def test_full(self, yaml_ast):
+    def test_full(self, yaml_ast, tmp_path):
         ServiceTemplate.parse(yaml_ast(
             """
             tosca_definitions_version: tosca_simple_yaml_1_3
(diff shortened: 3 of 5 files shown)


src/opera/parser/tosca/v_1_3/service_template.py
```diff
@@ -56,25 +56,41 @@ def normalize(cls, yaml_node):
         }, yaml_node.loc)
 
     @classmethod
-    def parse(cls, yaml_node, base_path, csar_path):
+    def parse(cls, yaml_node, base_path, template_path, parsed_templates):
         service = super().parse(yaml_node)
-        service.visit("prefix_path", csar_path)
-        service.merge_imports(base_path)
-        return service
+        if template_path is None:
```
tadeboro, line 61: Under what conditions can this be `None`? This is puzzling me a bit.

tests/opera/parser/tosca/v_1_3/test_service_template.py
```diff
@@ -1,8 +1,12 @@
 import pytest
+import pathlib
 
 from opera.error import ParseError
 from opera.parser.tosca.v_1_3.service_template import ServiceTemplate
+from opera.parser.tosca.v_1_3.import_definition import ImportDefinition
```
tadeboro, line 6: Is this import really needed?

tests/opera/parser/tosca/v_1_3/test_service_template.py
```diff
@@ -1,8 +1,12 @@
...
 from opera.parser.yaml.node import Node
+from opera.parser.tosca.path import Path
+from opera.parser import tosca
```
tadeboro, line 9: Again, do we need those two imports?

tests/opera/parser/test_tosca.py
```diff
@@ -101,3 +101,41 @@ def test_load_from_csar_subfolder(self, tmp_path, yaml_text):
...
+        name = pathlib.PurePath("template.yaml")
+        (tmp_path / name).write_text(yaml_text(
+            """
+            tosca_definitions_version: tosca_simple_yaml_1_3
+            imports: [ template.yaml ]
+            """
+        ))
+        tosca.load(tmp_path, name)
+
+    def test_duplicate_imports(self, yaml_ast, yaml_text, tmp_path):
```
tadeboro, line 115: This test has the same name as the one above, which means that only one of them is executed by pytest.

tadeboro, line 115: `yaml_ast` fixture is not needed in this test. Please remove it.

tests/opera/parser/test_tosca.py
```diff
@@ -101,3 +101,41 @@ def test_load_from_csar_subfolder(self, tmp_path, yaml_text):
...
+        (tmp_path / "sub/folder/imp2.yaml").write_text(yaml_text(
+            """
+            tosca_definitions_version: tosca_simple_yaml_1_3
+            data_types:
+              my_type:
+                derived_from: tosca.datatypes.xml
+            """
+        ))
+
+        tosca.load(tmp_path, name)
```
tadeboro, line 141: This code does not test anything new because all imports are still at the same level. Besides, no file is imported twice or more times here.

I would suggest a test similar to the one below:

```
def test_imports_from_multiple_levels(self, yaml_text, tmp_path):
    name = pathlib.PurePath("template.yaml")
    (tmp_path / name).write_text(yaml_text(
        """
        imports:
          - subfolder/a.yaml
          - subfolder/b.yaml
        """
    ))
    (tmp_path / "subfolder").mkdir()
    (tmp_path / "subfolder/a.yaml").write_text(yaml_text(
        """
        imports:
          - b.yaml
        """
    ))
    (tmp_path / "subfolder/b.yaml").write_text(yaml_text(
        """
        data_types:
          my_type:
            derived_from: tosca.datatypes.xml
        """
    ))

    tosca.load(tmp_path, name)
```

src/opera/parser/tosca/v_1_3/service_template.py
anzoman: Well yes this can be annoying but previously this parameter was named `csar_path` and I think that parameter could be None. For example look here: https://github.com/xlab-si/xopera-opera/blob/master/tests/opera/parser/tosca/v_1_3/test_service_template.py#L51, where the two parameters passed to the parse method are None.

Now I had to add additional parameter for `parsed_templates` to the tests and I used an empty set for that (set()). But still other params like `base_path` and `template_name` are None when testing service template: https://github.com/xlab-si/xopera-opera/blob/import-once/tests/opera/parser/tosca/v_1_3/test_service_template.py#L51

tests/opera/parser/tosca/v_1_3/test_service_template.py
anzoman: This was removed.

tests/opera/parser/tosca/v_1_3/test_service_template.py
anzoman: No, we don't. Imports were optimized and now it should be ok.

tests/opera/parser/test_tosca.py
anzoman: Actually the names have differed because  the first was `test_duplicate_import` and the second was `test_duplicate_imports` with plural for imports at the end. But this its not important anymore because this test was removed.

anzoman: Ok, this was removed.

tests/opera/parser/test_tosca.py
anzoman: I have included the proposed test. But now as expected the CircleCI pytests fail because the new problem emerged. It says `opera.error.ParseError: [MapWrapper] Duplicate keys 'my_type' found in subfolder/b.yaml:4:3 and subfolder/b.yaml:4:3` which was originally an exception that supported the idea about preventing duplicate keys within imports.

src/opera/parser/tosca/v_1_3/service_template.py
tadeboro: Well, you know what this means? That we need to fix the tests because I was lazy the first time around ;) The themplate path should aways contain something sensible.

tests/opera/parser/test_tosca.py
tadeboro: They did not differ enough then ;) My bad.

src/opera/parser/tosca/v_1_3/service_template.py
anzoman: Ok, so now I used `tmp_path` for both `base_path` and `template_name` in tests to prevent them from being None.

tests/opera/parser/test_tosca.py
tadeboro: This should work, which means that our "do not import things twice" implementation is bad. The general idea here is that loader works this way:

 1. Load `template.yaml` and add it to the set of processed files.
 2. Load `subfolder/a.yaml` and add it to the set of processed files.
 3. Load `subfolder/b.yaml` (from `subfolder/a.yaml`) and add it to the set of processed files.
 4. Skip loading `subfolder/b.yaml` (from `template.yaml`) because we already loaded that file.

cankarm:
Thank you for working on this!

I was thinking something - would it be good to throw some warning if this duplicates occur?

In some very rare cases user may bring together different models that use the template with the same name and different content or the templates are actually the same, just the versions are different. I'm not sure if this could happen in reality, but it would be very painful if this can happen and we do not have any output that this is possible. That means that xOpera would use only first import of the template all others (even if more appropriate) would be ignored. 

Currently we are making "name key check". Would be sum check suitable? At least for warning not for keeping two different same templates at the same time.

What do you think?

tests/opera/parser/test_tosca.py
anzoman: The bug was found and this was resolved so that all tests will pass.

tadeboro:
> I was thinking something - would it be good to throw some warning if this duplicates occur?

No.
 
> In some very rare cases user may bring together different models that use the template with the same name and different content or the templates are actually the same, just the versions are different. I'm not sure if this could happen in reality, but it would be very painful if this can happen and we do not have any output that this is possible. That means that xOpera would use only first import of the template all others (even if more appropriate) would be ignored.

We do not use template names to detect duplicates. We use relative paths to the template files from the CSAR root. And it is quite hard to put different content on the same path since opera does not accept symlinks.

For example, let us assume we have three files (I am only keeping the imports in for brevity):

```
# service.yaml
imports:
  - subfolder/a.yaml
  - subfolder/b.yaml

# subfolder/a.yaml
imports:
  - b.yaml

# subfolder/b.yaml
imports:
  - a.yaml
```

This is how parser will load this stuff (stuff in brackets represent a set of already loaded files at the end of the current step):

 1. Load `service.yaml` (service.yaml)
 2. Load `subfolder/a.yaml` - first import in `service.yaml` (service.yaml, subfolder/a.yaml)
 3. Load `b.yaml` - first import in `subfolder/a.yaml` (service.yaml, subfolder/a.yaml, subfolder/b.yaml)
 4. Skip `a.yaml` - first import in `subfolder/b.yaml` because that file has already been loaded.
 5. Skip `subfolder/b.yaml` - second import in `service.yaml` because that file has already been loaded.

> Currently we are making "name key check". Would be sum check suitable? At least for warning not for keeping two different same templates at the same time.

As I said before, name (or, more correctly, path) check is in opera's case as reliable as a checksum since we do not allow symlinks in the CSAR.

PR #34, opened 2020-05-05
Add CI/CD Python 3.8 testing and PyPI deploy

anzoman:
It's time to update our CI/CD pipeline by introducing changes 
that enable testing of the opera pip package with Python 3.8.
The new addon is also the ability to deploy opera Python package to
PyPI. This part is semi-automated and is invoked only by pushing a new
tag while on master branch.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,6 +20,11 @@ executors:
       - image: circleci/python:3.7.4
     environment:
       PIPENV_VENV_IN_PROJECT: true
+  python38:
+    docker:
+      - image: circleci/python:3.8.2
+    environment:
+      PIPENV_VENV_IN_PROJECT: true
 
 jobs:
   test36:
@@ -30,6 +35,10 @@ jobs:
     executor: python37
     steps:
       - run_tests
+  test38:
+    executor: python38
+    steps:
+      - run_tests
 
 workflows:
   version: 2
@@ -37,3 +46,4 @@ workflows:
     jobs:
       - test36
       - test37
+      - test38
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,10 +4,30 @@ commands:
   run_tests:
     description: Run test suite
     steps:
+      - common_setup
+      - run: make test
+
+  deploy_to_test_pypi:
+    description: Publish opera package to Python Package Index (PyPI) for testing
+    steps:
+      - common_setup
+      - run: env
+      - run: make build
+      - run: make test_release
+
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -6,3 +6,21 @@ test:
...
 fix:
 	pipenv run pytest -x tests
+
+build:
+	pipenv install -d setuptools wheel
+	pipenv run python3 setup.py sdist bdist_wheel
+
+test_release:
+	pipenv install -d twine
+	pipenv run twine upload --repository testpypi --username __token__ --password ${PYPI_TEST_TOKEN} --non-interactive dist/*
```
tadeboro, line 16: This is bad because `make` will echo the command back, revealing the token. I would suggest that we use environment variables `TWINE_USERNAME` and `TWINE_PASSWORD` for this task.

anzoman: Ok, I added extra env vars for CircleCI jobs. But I think that currently these variables are masked so token is not revealed. Correct me If I'm wrong please.

.circleci/config.yml
```diff
@@ -30,10 +43,27 @@ jobs:
...
     jobs:
       - test36
       - test37
+      - test38
+  deploy:
+    jobs:
+      - pypi_deploy:
+          filters:
+            tags:
+              only: /.*/
```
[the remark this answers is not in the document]
anzoman: I think that .* only tells that tag should have its parts separated by periods- So yes this is a good idea and I added this as you suggested.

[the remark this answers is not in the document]
tadeboro: No, `/.*/` is a regular expression that matches anything, which is bad. `.` in a regex means "any character" and the `*` means "zero or more repetitions of what is in front of me".

This might come in handy: https://regex101.com/r/ZiTlcd/1

Makefile
tadeboro: They are masked in the output on the web interface, but we should not rely on that. I would take things even further and advise you to NEVER put any kind of secret on a command-line because that just builds bad habits.

.circleci/config.yml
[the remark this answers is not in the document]
anzoman: I might have forgotten a meaning of this regex so thanks a lot for that.

Makefile
anzoman: Thanks for clearifying things a little bit more. But can you really always avoid placing secrets on a command-line in CI/CD jobs? I added `TWINE_USERNAME` and `TWINE_PASSWORD` env vars but this seems to have no effect as username is still visible in CircleCI pipline logs.

PR #36, opened 2020-05-15
Add first integration tests and daily/nightly xOpera CI/CD package testing

anzoman:
With these additions we populate our CI/CD configuration with daily
testing of xOpera python package. This new testing routine will now
occur every night at 3 o'clock and will help us to be sure that the
latest version of the orchestrator is working properly.

CircleCI will trigger a job where the most important opera commands
will be tested. In the very beginning of this worklflow the last
version of opera pip package is installed. This is followed by opera
validate operation and the deployment which is performed on testing
CSAR that includes most common TOSCA definitions. And last but not
least, after the deployment we also look for outputs and later on use
the undeploy command to demolish the deployed solution.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,6 +30,16 @@ commands:
       - run: make build
       - run: make release
 
+  create_and_push_stable_branch:
+    description: Create and push GitHub stable branch which points to the latest relased tag/commit
+    steps:
+      - add_ssh_keys:
+          fingerprints:
+            - "b6:94:61:c3:e1:8f:25:e5:43:6d:51:9e:43:c4:16:11"
+      - checkout
+      - run: git branch -f stable HEAD
+      - run: git push origin stable
+
 executors:
   python36:
     docker:
@@ -68,6 +78,10 @@ jobs:
     executor: python37
     steps:
       - deploy_to_production_pypi
+  push_stable_relase_branch:
+    executor: python37
+    steps:
+      - create_and_push_stable_branch
 
 workflows:
   version: 2
@@ -95,3 +109,6 @@ workflows:
               only: /[0-9]+\.[0-9]+\.[0-9]+/
             branches:
               ignore: /.*/
+      - push_stable_relase_branch:
+          requires:
+            - pypi_production_deploy
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,11 +1,19 @@
 version: "2.1"
 
 commands:
+  common_setup:
+    description: Prepare the repository for work
+    steps:
+      - checkout
+      - run: sudo pip install pipenv
+      - run: make init
+
   run_tests:
     description: Run test suite
     steps:
       - common_setup
       - run: make unit_test
+      - run: make integration_test
 
   deploy_to_test_pypi:
     description: Publish opera package to Python Package Index (PyPI) for testing
@@ -21,13 +29,6 @@ commands:
       - common_setup
       - run: make build
       - run: make release
-
-  common_setup:
-    description: Prepare the repository for work
-    steps:
-      - checkout
-      - run: sudo pip install pipenv
-      - run: make init
 
 executors:
   python36:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,6 +15,15 @@ commands:
       - run: make unit_test
       - run: make integration_test
 
+  run_integration_tests_using_released_opera_version:
+    description: Test xOpera CLI commands with testing service template
+    steps:
+      - checkout
+      - run: pip install --user --upgrade pip
+      - run: pip install --user wheel
+      - run: pip install --user opera
+      - run: OPERA=opera make integration_test
+
   deploy_to_test_pypi:
     description: Publish opera package to Python Package Index (PyPI) for testing
     steps:
@@ -70,6 +79,10 @@ jobs:
     executor: python38
     steps:
       - run_tests
+  nightly_xopera_test:
+    executor: python37
+    steps:
+      - run_integration_tests_using_released_opera_version
   pypi_test_deploy:
     executor: python37
     steps:
@@ -112,3 +125,13 @@ workflows:
       - push_stable_relase_branch:
           requires:
             - pypi_production_deploy
+
+  nightly_test:
+    triggers:
+      - schedule:
+          cron: "0 3 * * *"
+          filters:
+            branches:
+              only: stable
+    jobs:
+      - nightly_xopera_test
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,13 @@
+# List of all integration test scenarios
+integration_test_scenarios := $(wildcard tests/integration/*/runme.sh)
+
+# Default opera executable (tailored for developers). Can be overridden by
+# setting the OPERA environment variable to something else.
+OPERA ?= "pipenv run opera"
+
+# Point pipenv to the right file since we can execute tests from a subdir.
+export PIPENV_PIPFILE := $(realpath Pipfile)
+
 .PHONY: init
 init:
 	pipenv install -d
@@ -9,6 +19,13 @@ unit_test:
 .PHONY: fix
 fix:
 	pipenv run pytest -x tests
+
+.PHONY: integration_test
+integration_test: $(integration_test_scenarios)
+
+.PHONY: $(integration_test_scenarios)
+$(integration_test_scenarios):
+	cd $(dir $@) && bash $(notdir $@) $(OPERA)
 
 .PHONY: build
 build:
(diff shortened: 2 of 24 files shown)


anzoman:
@tadeboro this sounds great!

As you proposed, I moved our very first integration test to the suitable subdir. As for the CID/CD testing - I added the `test_integration` entry to out Makefile and used it in CircleCI `config.yml` within the `run_tests` job so that now unit tests as well as integration test are executed every time. I've noticed in the previous PR you added `fix` make command which I used here. And if I stick to the plan and your suggestions I also make nightly test in a way that it will loop through integration test folder and run every test in it using a special `runme.sh` script that receives opera executable path as the first argument and tries to run all the main opera commands (i.e. validate, deploy, undeploy, outputs).

I believe that there might still be some room for improvement but what do you think?

matejart:
May I suggest we use `*.yaml` instead of `*.yml`?

anzoman:
@matejart thanks  a lot for the suggestion. I see that `*.yaml` is being used everywhere else in xOpera so to be consistent I will update those testing artefacts to use the same ending.

.circleci/config.yml
```diff
@@ -22,12 +31,21 @@ commands:
...
+      - run: pip install --user wheel
+      - run: pip install --user opera
+      - run:
+          command: |
+            cd tests/integration
+            for integration_test in *; do
+              cd "${integration_test}"
+              sh runme.sh opera
+              cd ..
+            done
```
tadeboro, line 48: We should be able to reuse the make definition for this. See Makefile's suggestions below.

Makefile
```diff
@@ -10,6 +10,10 @@ unit_test:
 fix:
 	pipenv run pytest -x tests
 
+.PHONY: integration_test
+integration_test:
+	cd tests/integration && export PIPENV_PIPFILE=../../../Pipfile && for integration_test in *; do cd "$${integration_test}"; pipenv run sh runme.sh opera; cd ..; done
```
tadeboro, line 15: This line has quite a few problems ;)

Now, every time we see the loop in the Makefile, we should think again about our implementation because make works best if we only specify targets and their dependencies, and then let make do the looping. My suggestion would be the following:

```
# List of all integration test scenarios
integration_test_scenarios := $(wildcard tests/integration/*/runme.sh)

# Default opera executable (tailored for developers). Can be overridden by
# setting the OPERA environment variable to something else.
OPERA ?= "pipenv run opera"

# Poitn pipenv to the right file since we can execute tests from a subdir.
export PIPENV_PIPFILE := $(realpath Pipfile)

.PHONY: integration_test
integration_test: $(integration_test_scenarios)

.PHONY: $(integration_test_scenarios)
$(integration_test_scenarios):
	cd $(dir $@) && bash $(notdir $@) $(OPERA)
```

Implementing things this way also removes the hardcoded test command that we had in the Makefile before.

tests/integration/misc-tosca-types/runme.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
```
tadeboro, line 1: We should be more strict when running this command. I would suggest that we change this to

    #!/bin/bash
    set -euo pipefail

This makes sure that the script will terminate after the first failed command. Because doing undeploy when messed-up the code for deploy makes little sense.

tests/integration/misc-tosca-types/runme.sh
```diff
@@ -0,0 +1,15 @@
...
+
+if [ $# -ne 1 ]; then
+    echo "No arguments were supplied!"
+    exit 1
+fi
```
tadeboro, line 6: This is not needed when we use `set -u` at the beggining of the script.

.circleci/config.yml
```diff
@@ -1,11 +1,20 @@
...
+      - checkout
+      - run: sudo pip install pipenv
+      - run: make init
+
   run_tests:
     description: Run test suite
     steps:
       - common_setup
       - run: make unit_test
+      - run: make fix
```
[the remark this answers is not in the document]
anzoman: Ok, I removed that.

.circleci/config.yml
```diff
@@ -22,12 +31,21 @@ commands:
       - run: make build
       - run: make release
 
-  common_setup:
-    description: Prepare the repository for work
+  perform_opera_nightly_integration_testing:
+    description: Test xOpera CLI commands with testing service template
```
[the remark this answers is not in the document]
anzoman: Oh I see and will update this name.

.circleci/config.yml
anzoman: Ok, I will use `make integration_test` here. I assume that I will need to export `OPERA` env var here to use the opera installed from PyPI and not the one from pipenv. Is that right?

Makefile
anzoman: Yes as I thought my command is not the best thing here. An wow, your proposed solution seems really cool and promising so I will definitely include this.

tests/integration/misc-tosca-types/runme.sh
anzoman: Very good point, I will use the suggested changes.

tests/integration/misc-tosca-types/runme.sh
anzoman: Great! What was I thinking not to include this :) I will immediately remove the old part of the script and rely on just `set -u`.

.circleci/config.yml
```diff
@@ -22,12 +31,21 @@ commands:
...
     steps:
       - checkout
```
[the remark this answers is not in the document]
anzoman: Aha thanks for clarifying things here and also for Sensu Go example. I've noticed that opera already has one stable branch called `stable-0.1`.  Could we use that one or if not what would be the next step here and how would this be properly done for opera?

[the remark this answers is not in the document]
tadeboro: I would suggest that we create a new `stable` or `released` branch that always points for the latest released tag. Then we can follow the same procedure as we do in Sensu Go.

.circleci/config.yml
tadeboro: Exactly. Just provide `OPERA` environment variable via some CircleCI means and we should be gold.

.circleci/config.yml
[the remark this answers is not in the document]
anzoman: Ok, so if I udnerstand correctly we would have to maintain that `stable` branch to point to the lastest relased tag, right? So basically we would do a rebase onto master every time? As a replacement for CircleCI code checkout could we just use some run commands here to clone the opera repository and then just checkout to the latest tag somehow without creating a new separate branch?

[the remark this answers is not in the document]
tadeboro: In the git world, the "latest tag" is not well defined because it has no semantic meaning beyond "point to one and exactly one commit". Any other semantic meaning needs to be attached by humans.

The problem with the automatic determination of the latest tag is in the implementation details. Because git tags can have just about any name, the code that finds the latest stable one would need to be quite robust. But I do not mind adding it to the repo if we decide to go that way.

.circleci/config.yml
```diff
@@ -1,11 +1,29 @@
...
+
+  run_integration_tests_using_released_opera_version:
+    description: Test xOpera CLI commands with testing service template
+    steps:
+      - checkout
+      - run: pip install --user --upgrade pip
+      - run: pip install --user wheel
+      - run: pip install --user opera
+      - run: export OPERA=opera
+      - run: make integration_test
```
tadeboro, line 26: Are you sure this works? I never looked into the CircleCI internals, but I would guess that each run command is executed in a separate subprocess, making it impossible to define a variable in one command and use it in the next one. It would be safer just to do `run: OPERA=opera make integration_test` and avoid any pitfalls of the CI environment here.

Makefile
```diff
@@ -10,6 +10,24 @@ unit_test:
...
+# List of all integration test scenarios
+integration_test_scenarios := $(wildcard tests/integration/*/runme.sh)
+
+# Default opera executable (tailored for developers). Can be overridden by
+# setting the OPERA environment variable to something else.
+OPERA ?= "pipenv run opera"
+
+# Point pipenv to the right file since we can execute tests from a subdir.
+export PIPENV_PIPFILE := $(realpath Pipfile)
+
```
tadeboro, line 23: I would suggest that we move those line to the top of the Makefile in order to keep the configuration and target definitions separated a but more.

.circleci/config.yml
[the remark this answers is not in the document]
anzoman: I understand. So now I modified CircleCI config that it will create and push a new `stable` branch every time we create a new tag and deploy something to production. This way we won't need to maintain this manually and therefore our stable branch will always point to the latest pushed tag.

.circleci/config.yml
anzoman: Well well...I'm not so sure. I remember that with every `run` on a separate line a new shell is opened so I agree that we can join this two lines to avoid any possible trouble.

Makefile
anzoman: Yeah sure, this would be nicer.

Issue #41, opened 2020-05-18
Use case: Develop and test a TOSCA library


matejart:
## Description

This use case addresses the need of a developer of a [TOSCA library](https://github.com/xlab-si/xopera-opera/issues/41#issuecomment-630784044) to modify and test the library definitions and implementation with a relative convenience.

## Prerequisites

* The TOSCA library project initiated in, for example, `~/libraries/example`.
* A project, e.g., `~/projects/test-example` containing a service template to use the library.
* `opera` is installed in the actor's working environment.

## Steps

* Developer makes a symbolic link in the service template project to the TOSCA library project:

      $ cd ~/projects/test-example
      $ mkdir -p library
      $ ln -s ~/libraries/example library/example

* Developer initializes the deployment environment for the service in the editable mode and in the symlink-unsafe mode:

      $ opera init --allow-symlinks --editable .
      WARNING: You have enabled the use of symbolic links in the service template
      definition, which could lead to security issues. Please proceed responsibly!

* Developer runs the deploy of the service:

      $ opera deploy --inputs inputs.yaml --template service.yaml --allow-symlinks
      WARNING: You have enabled the use of symbolic links in the service template
      definition, which could lead to security issues. Please proceed responsibly!

## Expected results

* The service gets deployed according to the latest changes of the service template.
* The service gets deployed also using the latest changes of the imported TOSCA library.

anzoman:
I see that his issue raises a lot of new possibilities and also questions. But could you explain what exactly is this TOSCA library and what it's capable of doing? From my perspective I see it as a special kind of storage for service templates that can be connected to xOpera. And also - would this library be a part of opera itself or just an adjacent repository that will need to be used for these kind of deployments you posted here?

matejart:
@anzoman thank you for good questions.

I would define a TOSCA library as a redistributable and reusable collection of TOSCA definitions and corresponding interface operation implementations. It can provide a TOSCA profile for a specific domain. It is a fully distinct entity from opera. It can be included in an arbitrary service template and is usable in service deployment.

So the library would sit on a designated place on your hard drive. When packaging the CSAR, it gets incorporated into the CSAR.

anzoman:
@matejart thanks for the explanation and for making things much more clear to me. Also it's nice to see you opened the PR connected to this enhancement which will allow us to discuss things there even further and to jump into details.

anzoman:
[...]

Issue #39, opened 2020-05-18
Use case: Deploy a service template on a third-party platform


matejart:
Use case: Deploy a service template on a third-party platform
=============================================================

## Description

This use case covers a workflow of installing a [TOSCA library](https://github.com/xlab-si/xopera-opera/issues/41#issuecomment-630784044) for a third-party platform of choice, then using it in a service template to successfully run a deploy.

## Prerequisites

* A TOSCA library named `vendor.platform` exists and is available.
* A tool for managing TOSCA libraries exists and is installed in the actor's working environment. Let's call the tool `overture`.
* `opera` is installed in the actor's working environment.
* User's `$PWD` is in the root folder of their service template in a valid [CSAR-based structure](https://docs.oasis-open.org/tosca/TOSCA-Simple-Profile-YAML/v1.3/cs01/TOSCA-Simple-Profile-YAML-v1.3-cs01.html#_Toc20821079).

## Steps

* User installs the TOSCA library `vendor.platform`:

      $ overture install vendor.platform

* User authors a service template:

      $ cd ~/project/demo-034
      $ ls
      inputs.yaml service.yaml
      $ head -n 4 service.yaml
      tosca_definitions_version: tosca_simple_yaml_1_3

      imports:
        - /library/vendor/platform/main.yaml

* User creates a CSAR of the project:

      $ overture package --output demo-034.csar .

* User initializes the deployment environment for the service:

      $ opera init --inputs inputs.yaml demo-034.csar

* User runs the deploy of the service:

      $ opera deploy

## Expected results

* The TOSCA library `vendor.platform` is globally available for current and any subsequent deployments.
* The `demo-034.csar` is a TOSCA-compliant CSAR containing the service template and copies of any type definitions and imported implementations.
* An instance of the service as described in the service template is deployed.

anzoman:
Hi, according to this proposal implementing this with xOpera would mean that we will probably have to add some new features. Those thing could be:

- add `opera init` command 
- add `opera package` command
- add support for compressed CSARs and arhives (this enhancement is already present here: #27)   with `--template` option to tell opera which file in CSAR is the actual service template that will be executed
- refactor `opera deploy` command to start the deployment that was initialized in the deployment environment by the user

Does this sound reasonable?

matejart:
@anzoman yes, that's the implication. But first, let us see any issues or unforeseen implications of the use case itself.

dradX:
@matejart , @anzoman 

[...]

PR #42, opened 2020-05-19
Implementation of a packaging command

matejart:
This PR is an exploration into an `opera package` command, which creates a CSAR file from the service template and all the accompanying files in the root directory and the subtree of the service template YAML file. The initial implementation also includes a very simple attempt at importing [TOSCA libraries](https://github.com/xlab-si/xopera-opera/issues/41#issuecomment-630784044) into the CSAR, making the service template complete for use by `opera deploy`.

The PR includes an example where the packaging can be tested:

    $ cd opera/examples/library-test/service-template
    $ opera package -L .. -O /tmp/my-service.csar.zip service.yaml

Here are some initial takeaways:
* All service template and other TOSCA definitions need to import the library using the absolute path.
* The import path needs to match the subpath(s) in the local library repository.

This PR attemts to address a part of https://github.com/xlab-si/xopera-opera/issues/39.


examples/library-test/service-template/service.yaml
```diff
@@ -0,0 +1,11 @@
...
+
+imports:
+  - /library/steampunk.valves/main.yaml
+  - /library/steampunk.gears/main.yaml
+  - library/types.yaml
+
+topology_template:
+  node_templates:
+    dummy:
+      type: tosca.nodes.Root
```
anzoman, line 11: I just went through a lot of gears and valves here :smiley: I would propose this `topology_template` to be improved a little bit because now it uses just `dummy` node that derives from tosca root node. You may use gear and valve node TOSCA definitions you declared in other files that are already imported here.

My suggestion would be something like that:

```suggestion
topology_template:
  node_templates:
    dummy:
      type: tosca.nodes.Root
      
    steam_valve:
      type: steampunk.valves.Root
      
    landing_gear:
      type: steampunk.gears.Root

    my_big_fella:
      type: big_valve
```

What do you think?

matejart: Thank you, @anzoman , sounds good 👍

src/opera/commands/package.py
```diff
@@ -0,0 +1,71 @@
...
+        for path in source_path.glob("**/*"):
+            pathname = str(path)
+            rel_pathname = str(path.relative_to(source_path))
+            csar.write(pathname, arcname=rel_pathname)
+
+
+def package(args):
+    service_template_path = PurePath(args.service_template.name)
+    service_path = service_template_path.parent
+    library_path = PurePath(args.library_source)
```
anzoman, line 54: Would this go through if you run this command without `--library-source` or `-L` option or should there be some existence check? I tried to run something like `opera package service.yaml` and got this error `TypeError: expected str, bytes or os.PathLike object, not NoneType` on line 54.

src/opera/commands/package.py
```diff
@@ -0,0 +1,71 @@
...
+from opera.storage import Storage
+
+
+def add_parser(subparsers):
+    parser = subparsers.add_parser(
+        "package",
+        help="Package service template and all accompanying files into a CSAR"
+    )
+    parser.add_argument(
+        "--output", "-O", type=argparse.FileType("w"),
```
anzoman, line 18: Just wondering - could option `-o` be better than `-O` capitalized? And the same question applies to `-L`. Because till now we always used small letters for the command options.

anzoman:
@matejart it's really nice to see the examples and unit tests. I would also suggest to add an integration test where `opera package` command would be used to prepare the CSAR and then also `opera deploy` for the deployment. But on the other hand maybe it it too soon to add these and it would be better to add them when `opera init` command will be implemented.

src/opera/commands/package.py
cankarm: This could be added as a future wish/feature.

cankarm:
Conflicts in __init__ need to be resolved.

anzoman:
@cankarm I've rebased the branch onto master so merge conflicts are now gone.

anzoman:
I have reshaped the opera package command with my last two commits.

The current development process shows clearly that we don't need so 
complicated behaviour for the opera package command (e.g. tosca library 
is just an idea and is not yet implemented in opera) and that reaching its 
goal (i.e. compressed CSAR) can be simplified.

TOSCA CSARs contain the blueprint of the application that we wanna
orchestrate. The opera package CLI command can be used to create a
valid TOSCA CSAR by packaging together the TOSCA service template and
all accompanying files. In general, opera package receives a directory
(where user's TOSCA templates are located) and produces a compressed
CSAR file. The command can create the CSAR if there is at least one
TOSCA YAML file in the input folder. If the CSAR structure is already
present (if TOSCA-Metadata/TOSCA.meta exists and all other TOSCA CSAR
constraints are satisfied) the CSAR is created without an additional
temporary directory. And if not, the files are copied to the tempdir,
where the CSAR structure is created and at the end the tempdir is
compressed. The input folder is the mandatory positional argument, but
there are also other command flags that can be used:

  1. --service-template/-t is the name of the main TOSCA template
     from the input folder that will be used as CSAR entrypoint.
     When not set, the root YAML file from the folder is used.

  2. --output/-o is the CSAR file path output that will be created.
     If not set, the output CSAR file will be named using UUID.

  3. --format/-f is the option that specifies the compressed file
     format. Currently, there are two choices here: 'zip' or 'tar'
     with 'zip' being the default compression method.

Here's an example of how to use the new `opera package` CLI command:

```console
# the structure of the hello directory (from examples/hello):
(.venv) anzoman@ubuntu:~/Desktop/xopera-opera$ tree examples/hello
examples/hello
├── playbooks
│   ├── create.yaml
│   └── delete.yaml
└── service.yaml

1 directory, 3 files

# create a CSAR from hello directory
(.venv) anzoman@ubuntu:~/Desktop/xopera-opera$ opera package examples/hello/
CSAR was created and packed to '/home/anzoman/Desktop/xopera-opera/opera-package-28bd2e8d9d49450e8a6ece25dc9cca42.zip'.

# package hello directory to CSAR (also supply TOSCA service template name and CSAR output path)
(.venv) anzoman@ubuntu:~/Desktop/xopera-opera$ opera package -t service.yaml -o test.zip examples/hello/
CSAR was created and packed to '/home/anzoman/Desktop/xopera-opera/test.zip'.

# the tree of more complex directory with predefined CSAR structure (similar to tests/integration/misc-tosca-types/):
(.venv) anzoman@ubuntu:~/Desktop/xopera-opera$ tree templates/misc-tosca-types/
templates/misc-tosca-types/
├── modules
│   ├── artifact_types
│   │   └── test
│   │       └── test.yaml
│   ├── capability_types
│   │   └── test
│   │       └── test.yaml
│   ├── data_types
│   │   └── test
│   │       └── test.yaml
│   ├── group_types
│   │   └── test
│   │       └── test.yaml
│   ├── interface_types
│   │   └── test
│   │       └── test.yaml
│   ├── node_types
│   │   ├── file
│   │   │   ├── file.yaml
│   │   │   └── playbooks
│   │   │       ├── create.yaml
│   │   │       └── delete.yaml
│   │   ├── hello
│   │   │   ├── hello.yaml
│   │   │   └── playbooks
│   │   │       ├── create.yaml
│   │   │       └── files
│   │   │           └── test.txt
│   │   ├── noimpl
│   │   │   └── noimpl.yaml
│   │   ├── setter
│   │   │   ├── playbooks
│   │   │   │   └── create.yaml
│   │   │   └── setter.yaml
│   │   └── test
│   │       ├── file.test
│   │       ├── playbooks
│   │       │   └── create.yaml
│   │       └── test.yaml
│   ├── policy_types
│   │   └── test
│   │       └── test.yaml
│   └── relationship_types
│       └── test
│           ├── playbooks
│           │   └── test.yaml
│           └── test.yaml
├── runme.sh
├── service-template.yaml
└── TOSCA-Metadata
    └── TOSCA.meta

# create a TOSCA CSAR with the 'tar' compression format (also provide service template name and output CSAR path)
(.venv) anzoman@ubuntu:~/Desktop/xopera-opera$ opera package -t service-template.yaml -o test.tar -f tar templates/misc-tosca-types/
CSAR was created and packed to '/home/anzoman/Desktop/xopera-opera/test.tar'.
```

src/opera/parser/tosca/csar.py
```diff
@@ -1,19 +1,92 @@
...
+        try:
+            root_yaml_files = []
+            root_yaml_files.extend(Path(self._csar_name).glob('*.yaml'))
+            root_yaml_files.extend(Path(self._csar_name).glob('*.yml'))
+
+            if len(root_yaml_files) != 1:
+                raise ParseError("There should be one root level yaml "
+                                 "file in the root of the CSAR: {}."
+                                 .format(list(map(str, root_yaml_files))),
+                                 self)
```
sstanovnik, line 29: I'd suggest an additional parameter to `package` that lets you choose one of the (more than one) files, falling back to selecting this (only) one if that isn't specified. That would be in line with my expectation of `package` just mainly building the metadata file.

sstanovnik: Oh, I just noticed that this argument exists as `-t/--service-template`. The code below still fails if there is more than one yaml file at the root though.

src/opera/parser/tosca/csar.py
```diff
@@ -1,19 +1,92 @@
...
+            if meta_file_path.exists():
+                # check existing TOSCA.meta file
+                with meta_file_path.open() as meta_file:
+                    self._tosca_meta = yaml.safe_load(meta_file)
+
+                self._validate_csar_version()
+                self._validate_tosca_meta_file_version()
+                template_entry = self._get_entry_definitions()
+                self._get_created_by()
+                self._get_other_definitions()
```
sstanovnik, line 52: What does this do?

src/opera/parser/tosca/csar.py
```diff
@@ -1,19 +1,92 @@
...
+                if not Path(self._csar_name).joinpath(
+                        service_template).exists():
+                    raise ParseError('The supplied TOSCA service template '
+                                     'file "{}" does not exist in folder '
+                                     '"{}".'.format(service_template,
+                                                    self._csar_name), self)
+
+            meta_file_path = Path(
+                self._csar_name) / "TOSCA-Metadata" / "TOSCA.meta"
+            if meta_file_path.exists():
```
sstanovnik, line 43: If you get `--service-template` specified and a meta file with it set exists, you silently ignore that, right? I don't know whether you should error, warn, or just silently change the value in the meta file.

src/opera/parser/tosca/csar.py
```diff
@@ -1,19 +1,92 @@
...
+                    # copy all the needed CSAR files to tempdir
+                    shutil.copytree(self._csar_name, tempdir,
+                                    dirs_exist_ok=True)
+
+                    # create TOSCA-Metadata/TOSCA.meta file using the specified
+                    # TOSCA service template or directory root YAML file
+                    content = ("TOSCA-Meta-File-Version: 1.1\n"
+                               "CSAR-Version: 1.1\n"
+                               "Created-By: xOpera TOSCA orchestrator\n"
+                               "Entry-Definitions: {}".
```
sstanovnik, line 75:
```python3
you_can_use = """
multiline string
literals {0}
""".format("to make this a bit more readable")
```

src/opera/commands/package.py
```diff
@@ -0,0 +1,68 @@
...
+
+from opera.error import DataError, ParseError
+from opera.parser.tosca.csar import CloudServiceArchive
+
+
+def add_parser(subparsers):
+    parser = subparsers.add_parser(
+        "package",
+        help="Package service template and all accompanying files into a CSAR"
+    )
```
sstanovnik, line 14: Needs the verbose switch to be consistent with other commands.

src/opera/commands/package.py
```diff
@@ -0,0 +1,68 @@
...
+        raise argparse.ArgumentTypeError("Directory {} is not a valid path!"
+                                         .format(args.service_template_folder))
+
+    # if the output is set use it, if not create a random file name with UUID
+    if args.output:
+        # remove the file extension if needed
+        csar_output = str(Path(args.output).with_suffix(''))
+    else:
+        # use uuid4 to create a unique file name
+        csar_output = "opera-package-" + str(uuid.uuid4().hex)
```
sstanovnik, line 45: See my comments on UUIDs in #125, the same thing applies here.

src/opera/commands/package.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def package(input_dir: str, csar_output: str, service_template: str,
+            csar_format: str) -> str:
+    """
+    :raises ParseError:
+    :raises DataError:
+    """
+    csar = CloudServiceArchive(input_dir)
+    return csar.package_csar(csar_output, service_template, csar_format)
```
sstanovnik, line 68: Does this return the generated filename? It doesn't make to return the exact thing you gave it in `csar_output`.

src/opera/commands/package.py
```diff
@@ -0,0 +1,68 @@
...
+             "file from the root of the input folder",
+    )
+    parser.add_argument(
+        "--output", "-o",
+        help="Output CSAR file path",
+    )
+    parser.add_argument(
+        "--format", "-f", choices=("zip", "tar"),
+        default="zip", help="CSAR compressed file format",
+    )
```
sstanovnik, line 27: Are tars conformant to the tosca standard?

sstanovnik (changes requested):

src/opera/parser/tosca/csar.py
anzoman: This creates a portal to the underworld. Just kidding. This gets the entries `Created-By` and `Other-Definitions` from the TOSCA meta file. The two keynames are not required in the meta file but just optional. However, we don't use the anywhere right now but they will probably be used in the future when (and if) opera evolves. So this is actually not needed here (the same applies for the CSAR validation where these two lines are present as well) but we can keep this anyways to be prepared for further implementation. I'm also ok if you want me to remove these.

src/opera/parser/tosca/csar.py
anzoman: Aha, do you mean the case when I would specify TOSCA service template path with `--service-template` to one file and then in the `TOSCA.meta` (within `Entry-Definitions`) it would be set to another different file path. Yes, this could be a problem as here I only take the path in `TOSCA.meta` and forget about the first one. This can be resolved by adding a check if the two entries match. If they don't I believe that an error should be thrown.

src/opera/commands/package.py
anzoman: Well, I don't know actually.

TOSCA standard says [here](https://docs.oasis-open.org/tosca/TOSCA/v1.0/os/TOSCA-v1.0-os.html#_Toc356403711): `A CSAR is a zip file containing at least two directories, the TOSCA-Metadata directory and the Definitions directory. `, but it doesn't prevent other compression formats explicitly. I think I saw some tar compressed CSARs in RADON but I'm not sure. And if I search a little bit more: [here](https://blueprints.launchpad.net/heat-translator/+spec/tosca-csar-translation), it is said that `TOSCA template files most likely will be provided as part of CSAR archive file (typically zip or tar).`, or if you look at [toscafy CSAR tool](https://github.com/toscafy/toscafy/blob/6fa7e24f75503cdc536aec3b97475f342fb21377/swagger.yaml#L69) it offers packing the CSAR files with zip or tar compression formats. It seems that [Aria TOSCA orchestrator](http://ariatosca.incubator.apache.org/docs/html/aria.cli.html#aria.cli.service_template_utils.get) also uses different compression formats. @matejart do you happen to know this?

src/opera/parser/tosca/csar.py
anzoman: Hmm yes we have an option (`-t/--service-template`) to point `opera package` command to the selected TOSCA service template. But this will still need some reshaping. According to [TOSCA v1.3](https://docs.oasis-open.org/tosca/TOSCA-Simple-Profile-YAML/v1.3/cos01/TOSCA-Simple-Profile-YAML-v1.3-cos01.html#_Toc26969474) there are two possible CSAR cases:

- CSAR with `TOSCA.meta`:  here you are allowed to have more than one YAML file in the root of the CSAR (because you already have a reference to the entrypoint in the `Entry-Definitions` keyword) 
- CSAR without `TOSCA.meta`: here you should have only one YAML file at the root of the CSAR that contains the `metadata` YAML map entry.

So, with `opera package` we are currently aiming to create the first type of the CSAR (with `TOSCA.meta`) rather than modifying the specified YAML file (we would need to add `metadata` entry to that file if we wanted to create the second CSAR type). At the moment we create a mixture between the two types as we create `TOSCA.meta` and we also allow only one root level yaml file in the root of the CSAR.

So there are many ways. We can modify current implementation so that we allow multiple root YAML files in the root of the CSAR - here if the user would specify the CSAR entrypoint with `-t/--service-template` it would be ok if there would be multiple YAML files in the root of the CSAR, and if he omits that CLI flag the packager would check if there is only one file in the root to select it as the new entrypoint or if there are multiple it would raise a ParseError like it does now.

Or we can add an additional switch (for instance `--without-meta-file`) for user to decide which CSAR type he would like to receive. What do you propose?

anzoman: Another question: Is `-t/--service-template``okay? Or does `-e/--csar-entrypoint` sound better?

src/opera/parser/tosca/csar.py
anzoman: Done.

src/opera/commands/package.py
anzoman: Dunzo.

src/opera/parser/tosca/csar.py
anzoman: A said I've added check for that:

```python
                if service_template and template_entry != service_template:
                    raise ParseError('The file entry "{}" defined within '
                                     '"Entry-Definitions" in '
                                     '"TOSCA-Metadata/TOSCA.meta" does not '
                                     'match with the file name "{}" supplied '
                                     'in service_template CLI argument.'.
                                     format(template_entry, service_template),
                                     self)
```

src/opera/commands/package.py
anzoman: Yes, this returns the name of the created archive file. If the desired name was specified with `-o/--output` the result should be almost the same as `csar_output` param (almost because I strip off the the file extension before). If not then this returns the name of the created CSAR file that was generated with UUID.

src/opera/commands/package.py
anzoman: Yes sir!

anzoman:
@sstanovnik package and unpackage commands are very connected so the source branch for this PR (`packager`) was rebased onto `unpackager` branch that has to be merged first in this PR: #125.

src/opera/commands/package.py
anzoman: This has been tackled with the same [`generate_random_pathname`](https://github.com/xlab-si/xopera-opera/pull/125#discussion_r540091619) function that was introduced in the connected PR for opera package.

src/opera/parser/tosca/csar.py
anzoman: As we talked, its currently better to create just the first type of the CSAR with `TOSCA.meta` and keep the `-t/--service-template` option.

src/opera/parser/tosca/csar.py
anzoman: I will remove those two (`_get_created_by` and `_get_other_definitions`) method calls because they do nothing. The same stands for `_get_author`. I've also added the comments that these functions are not called anywhere but are prepared for future uses.

src/opera/commands/package.py
anzoman: We decided to keep both zip or tar CSAR formats because TOSCA doesn't specify any exact mimetype here.

src/opera/commands/package.py
anzoman: We decided to add some more logic for file extensions. Now according to format, the `zip` or `tar` extension is added if the desired output string doesn't have it.

anzoman: Because `shutil.make_archive` already adds the extension according to format (.zip or .tar) we now had to remove if if already exists.

sstanovnik (approved):

sstanovnik:
This looks nice now, thanks!

PR #44, opened 2020-05-28
Update the hello world README example

anzoman:
These changes include some updates for opera's README and more
accurately here we are providing the missing information for the
quickstart section where (although it was evident) it is now specified
that user has to clone the opera's GitHub repository before deploying
the examples within.

cc @cankarm

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -17,10 +17,11 @@ The simplest way to test ``opera`` is to install it into virtual environment::
   $ python3 -m venv .venv && . .venv/bin/activate
   (.venv) $ pip install opera
 
-To test if everything is working as expected, we can now try to deploy a
+To test if everything is working as expected, we can now clone xOpera's GitHub repository and try to deploy a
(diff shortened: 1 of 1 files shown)


cankarm:
would it be better to add a command instead to do in text?

git clone <repo>?

anzoman:
@cankarm maybe you were seduced by commit message that does not explicitly tell that command instructions were also updated. So there actually is a `git clone` command included and the example now says that user has to do the following to be able to run the hello world example:

```console
  (.venv) $ git clone git@github.com:xlab-si/xopera-opera.git
  (.venv) $ cd xopera-opera/examples/hello
  (.venv) $ opera deploy service.yaml
```

cankarm:
I'm sorry, @anzoman. Actually I opened the changes, saw the text change and ignored the as started with `cd ....`. instead of expected `clone`, and thought that this red line is green :D

PR #45, opened 2020-05-28
Support for attribute mapping in operations

matejart:
*TOSCA Simple Profile in YAML Version 1.3* specifies in Section 3.6.17, that an operation may include keyname `outputs`, which is a map of attribute mappings. This then interpreted as *an optional map of attribute mappings that specify named operation output values and their mappings onto attributes of the node_type or relationship that contains the interface within which the operation is defined.*
There is an example of usage in Section 2.16.1.

This PR implements the support for this feature. We are now able in each operation to specify a list of key names that the operation's implementation returns at the end of its running. For each key, we can provide a mapping to an attribute that will receive the new value. We support referencing attributes of the current node template's instance in node template interfaces (`SELF`), and referencing the target (`TARGET`) or the source (`SOURCE`) of a relationship as well as the relationship's instance (`SELF`) as well.

The enclosed example illustrates the use of resource mapping in an example, where we have one node of type A that depends on multiple nodes of type B. For its operation, each instance of type A needs to keep a list of the dependencies' of all instances of the dependent type B instance.

### DIFF
--- a/examples/attribute_mapping/playbooks/create-student.yaml
+++ b/examples/attribute_mapping/playbooks/create-student.yaml
@@ -1,0 +1,10 @@
+---
+- hosts: all
+  gather_facts: false
+
+  tasks:
+
+    - name: Set attributes
+      set_stats:
+        data:
+          student_id: "student-{{ id }}"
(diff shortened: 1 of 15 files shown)


anzoman:
Hi,

great, this looks promising and it definitely is a useful new feature. I tried to understand your changes and also to build opera and run the new example CSAR locally. Eventually everything worked fine and at the end the teacher managed to gather all the students ;).

PR #48, opened 2020-06-01
Implement the ability to relocate .opera storage folder

anzoman:
Within this update we are improving some of opera's commands by adding
the -l (or --location) command line option that brings the possibility
to control the location where opera will store the information about
orchestration and deployment. So with these changes user can now
specify the location of the .opera folder along with opera deploy,
undeploy and outputs commands.

Hitherto xOpera created the aforementioned .opera storage folder in the
current working directory which resulted in the lack of flexibility
when deploying the (same) solutions multiple times. Now there is a
chance to use use the same deployment CSARs with several executions of
opera without any interference. Users will be able to relocate and store
orchestration info to the suitable locations which will also make
those things reusable and easier for testing the deployment.

I've tested the changes locally and have run the following commands on integration test CSAR example:

```console
# deploy
opera deploy -l src/ -i tests/integration/misc-tosca-types/inputs.yaml tests/integration/misc-tosca-types/service-template.yaml
# udeploy
opera outputs --location src/
# outputs
opera undeploy --location /src
```

There might still be some room for improvement (if this gets approved) like:

- adding unit tests for the new option
- considering the possibility to add `--location` option as a global parameter

cc @sstanovnik this could be useful for your SaaS API/xOpera API

### DIFF
--- a/src/opera/commands/deploy.py
+++ b/src/opera/commands/deploy.py
@@ -6,12 +6,17 @@ import yaml
 from opera.error import DataError, ParseError
 from opera.parser import tosca
 from opera.storage import Storage
+from os import path
 
 
 def add_parser(subparsers):
     parser = subparsers.add_parser(
         "deploy",
         help="Deploy service template from CSAR"
+    )
+    parser.add_argument(
+        "--instance-path", "-p",
+        help=".opera storage folder location"
     )
     parser.add_argument(
         "--inputs", "-i", type=argparse.FileType("r"),
@@ -24,7 +29,10 @@ def add_parser(subparsers):
 
 
 def deploy(args):
-    storage = Storage(Path(".opera"))
+    if args.instance_path and not path.isdir(args.instance_path):
+        raise argparse.ArgumentTypeError("Directory {0} is not a valid path!".format(args.instance_path))
+
+    storage = Storage(Path(args.instance_path).joinpath(".opera")) if args.instance_path else Storage(Path(".opera"))
     storage.write(args.csar.name, "root_file")
 
     # TODO(@tadeboro): This should be part of the init command that we do not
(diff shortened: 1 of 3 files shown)


sstanovnik:
_Nice_, this will come in really handy.

I'd suggest adding `--location` to be a global parameter. It's definitely useless for `validate`, but maybe other commands down the line?

Also, I'd prefer not appending `.opera/` automatically, instead keeping `$PWD/.opera/` as the default and being able to have a flatter structure. Maybe `--location` could be `--storage`? `--state`? Otherwise it could get confusing as to whether this is

* the location of the CSAR,
* the location of the service template,
* the location which is the root of relative paths within templates,
* the (location of the) CWD where implementations get executed or
* the location of the state storage (ding ding ding!).

Well, just some things to think about. I could use what you've implemented until now as a solution to my woes and it would work perfectly :)

matejart:
Thank you, @anzoman , this is a good enhancement that will come useful. Even though it does place some additional burden on the user to keep track which location was in use for which particular deploy, it does make things more flexible. At some point down the line, we should consider also supporting an environment variable to keep this information in.

I agree with @sstanovnik on the point of "location" being ambiguous, so perhaps something along the lines of an "instance path" would make things clearer. And I too would be fine with keeping `$PWD/.opera` as a default and not append it automatically.

anzoman:
@sstanovnik and @matejart thanks for your comments and suggestions.

I have made some changes. Amongst them I renamed the fuzzy `--location` option to `--instance-path` (or just `-p`) and I also  brought back the default `$PWD/.opera` location which is used when `--instance-path` was not explicitly set rather than setting this option's default value to current dir  with getcwd and then  automatically appending opera to it.

@matejart do you also think that this new option should be a global CLI parameter? Probably then I would need to modify the cli python file (in `src/opera/cli.py`), add this argument to the main argument parser and then also providing the main parser's name to a `parents` list argparse parameter for all the subparsers.

PR #52, opened 2020-06-03
added topology template merge

### LINKED ISSUE #49: node_templates section defined in imported service templates
## Description
In some cases, it would be useful to import a service template that already contains some node_templates defined. For example, in a situation, when a service template describing existing infastruture that cannot be modified externally (e.g. HPC nodes) is automatically generated by some service and later imported into other templates.  
## Prerequisites
- `opera` installed
- A valid TOSCA Service Template file `service.yaml` that contains an import definition of another valid TOSCA Service Template file with **topology_template** and **node_templates** sections defined.
## Steps
- User runs deploy `opera deploy service.yaml`
## Current behaviour
- xOpera returns an error `Duplicate keys 'node_templates' found in service.yml`
## Expected behavior
- Service template is deployed without error.

alexmaslenn:
Closes #49 

Merge topology template the same way we merge service template. Run through the Map and merge all sections.

### DIFF
--- a/src/opera/parser/tosca/v_1_3/topology_template.py
+++ b/src/opera/parser/tosca/v_1_3/topology_template.py
@@ -64,3 +64,20 @@ class TopologyTemplate(Entity):
                 ),
             ) for name, definition in self.get("outputs", {}).items()
         }
+
+    def merge(self, other):
+        for key in (
+                "inputs",
+                "node_templates",
+                "data_types",
+                "relationship_templates",
+                "groups",
+                "policies",
+                "outputs"
+        ):
+            if key not in other.data:
+                continue
+            if key in self.data:
+                self.data[key].merge(other.data[key])
+            else:
+                self.data[key] = other.data[key]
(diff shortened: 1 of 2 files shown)


src/opera/parser/tosca/v_1_3/topology_template.py
```diff
@@ -64,3 +64,22 @@ def collect_outputs(self, service_ast):
                 ),
             ) for name, definition in self.get("outputs", {}).items()
         }
+
+    def merge(self, other):
+        # TODO(@alexmaslenn): Should we merge all topology template sections
+        # or should we take some of them only from root template (e.g outputs)
```
matejart, line 70: For me, the implementation is fine, so I suggest removing this comment.

alexmaslenn:
@matejart @anzoman updated commits and added tests as you had asked.

matejart:
Thank you, @alexmaslenn , for having contributed the tests as well!

And thank you, @anzoman , for making this correct call so that the change could be included without waiting for my response.

Issue #32, opened 2020-04-20
Add support for policy targets and triggers


anzoman:
Currently it is possible to declare TOSCA policy definition inside the `topology_template` section of the TOSCA service template but there is no support for policy targets and triggers so I thought that I'd be nice to have them and to be able to define a policy like in the example below.

```yaml
<policy_name>:
  type: <policy_type_name>
  description: <policy_description>
  metadata:
    <map of string>
  properties:
    <property_assignments>
  targets: [<list_of_policy_targets>]
  triggers:
    <trigger_definitions>
```

anzoman:
We are getting closer! Parsing of policy targets was supported with #54 and TOSCA triggers were introduced in #75. All we have to do now is to figure out how we're gonna treat those policies and invoke triggers with opera. And because this is not so trivial, we will need to discuss thing further. So I would like to call upon you - @cankarm, @sstanovnik, @matejart, @tadeboro, @dradx, @alexmaslenn @MysteriousWolf - if any of you has an idea how this should work within opera, please share your opinions here.

anzoman:
Connected with opera scaling issue #108.

anzoman:
With #187 and  #188 TOSCA policies (and triggers) will be supported not just for parsing, but also during the orchestration.

PR #54, opened 2020-06-05
Support targets keyname in policy type definitions

anzoman:
In this commit we are updating our TOSCA parser with the support for
policy type targets which is an optional list of valid node types or
group types the policy type can be applied to. When used is
node_templates section of the main TOSCA service templates policy
targets represent the optional list of names of node templates or
groups that the actual policy gets applied to. Contents of this
commit are also linked to the issue #32. The only thing that still
needs to be implemented are policy type triggers.


anzoman:
@matejart I've tested this again and updated the content. The only problem in the execution was when having the same entity name present in `node_templates` and in `groups`. Now opera raises an exception if this scenario carries out.

Issue #27, opened 2020-03-09
Support for compressed CSARs


cankarm:
Current release supports CSARs, but not the compressed version of CSAR.

anzoman:
This issue was addressed in #55. Now opera supports deploying/undeploying service templates and also the compressed CSAR files. This was achieved by implementing `opera init` command and the deployment can be done as follows:

```console
# initialize the CSAR (from the prepared compressed CSAR file) 
# also provide inputs
opera init --inputs inputs.yaml

# deploy the initialized CSAR
opera deploy

# uddeploy the deployed CSAR
opera deploy
```

So I guess that this is it - we can close the issue!

PR #55, opened 2020-06-09
Incorporate the support for compressed CSAR files

anzoman:
Up until now opera supported deploying uncompressed CSARs, but not the
compressed version of CSAR file. Here we accomplished that opera will
recognize also zip compressed TOSCA template files and will allow to
either deploy or undeploy the service template from CSAR. These
changes are linked to the issue #27 that provided the enhancement.

I have tested these changes locally with:

```console
# deploy compressed test.csar
opera deploy -i inputs.yaml -t service-template.yaml test.csar

# udeploy the compressed CSAR
opera undeploy -c test.csar
```

If needed I can also provide an integration test with compressed CSAR and those new CLI options.

cc @cankarm  just mentioning you because it was you who filed this issue and because this was one of the important requirements for the the RADON project.

### DIFF
--- a/src/opera/commands/deploy.py
+++ b/src/opera/commands/deploy.py
@@ -39,7 +39,7 @@ def deploy(args):
     if args.instance_path and not path.isdir(args.instance_path):
         raise argparse.ArgumentTypeError("Directory {0} is not a valid path!".format(args.instance_path))
 
-    storage = Storage(Path(args.instance_path) / ".opera") if args.instance_path else Storage(Path(".opera"))
+    storage = Storage(Path(args.instance_path)) if args.instance_path else Storage(Path(".opera"))
 
     if args.workers < 1:
         print("{0} is not a positive number!".format(args.workers))
--- a/src/opera/commands/deploy.py
+++ b/src/opera/commands/deploy.py
@@ -20,7 +20,7 @@ def add_parser(subparsers):
     )
     parser.add_argument(
         "--inputs", "-i", type=argparse.FileType("r"),
-        help="YAML file with inputs",
+        help="Optional: YAML file with inputs to override the inputs supplied in init",
     )
     parser.add_argument(
         "--workers", "-w",
@@ -28,9 +28,10 @@ def add_parser(subparsers):
             threads (positive number, default 1)",
         type=int, default=1
     )
-    parser.add_argument("csar",
-                        type=argparse.FileType("r"),
-                        help="cloud service archive file")
+    parser.add_argument("template",
+                        type=argparse.FileType("r"), nargs='?',
+                        help="TOSCA YAML service template file",
+                        )
     parser.set_defaults(func=deploy)
 
 
@@ -38,8 +39,7 @@ def deploy(args):
     if args.instance_path and not path.isdir(args.instance_path):
         raise argparse.ArgumentTypeError("Directory {0} is not a valid path!".format(args.instance_path))
 
-    storage = Storage(Path(args.instance_path).joinpath(".opera")) if args.instance_path else Storage(Path(".opera"))
-    storage.write(args.csar.name, "root_file")
+    storage = Storage(Path(args.instance_path) / ".opera") if args.instance_path else Storage(Path(".opera"))
 
     if args.workers < 1:
         print("{0} is not a positive number!".format(args.workers))
@@ -47,15 +47,33 @@ def deploy(args):
 
     # TODO(@tadeboro): This should be part of the init command that we do not
     # have yet.
+    if args.template:
+        storage.write(args.template.name, "root_file")
+        service_template = args.template.name
+    else:
+        if storage.exists("root_file"):
+            service_template = storage.read("root_file")
+        else:
+            print("CSAR or template root file does not exist. Maybe you have forgotten to initialize it.")
+            return 1
+
     try:
-        inputs = yaml.safe_load(args.inputs) if args.inputs else {}
-        storage.write_json(inputs, "inputs")
+        if storage.exists("inputs"):
+            inputs = yaml.safe_load(storage.read("inputs"))
+        else:
+            inputs = {}
+            storage.write_json(inputs, "inputs")
+
+        if args.inputs:
+            inputs = yaml.safe_load(args.inputs)
+            storage.write_json(inputs, "inputs")
+
     except Exception as e:
         print("Invalid inputs: {}".format(e))
         return 1
 
     try:
-        ast = tosca.load(Path.cwd(), PurePath(args.csar.name))
+        ast = tosca.load(Path.cwd(), PurePath(service_template))
         template = ast.get_template(inputs)
         topology = template.instantiate(storage)
         topology.deploy(args.workers)
--- a/src/opera/commands/init.py
+++ b/src/opera/commands/init.py
@@ -1,0 +1,107 @@
+import argparse
+from os import path
+from pathlib import Path, PurePath
+from zipfile import ZipFile, is_zipfile
+
+import yaml
+
+from opera.error import DataError, ParseError
+from opera.parser import tosca
+from opera.storage import Storage
+from opera.parser.tosca.csar import CloudServiceArchive
+
+
+def add_parser(subparsers):
+    parser = subparsers.add_parser(
+        "init",
+        help="Initialize the deployment environment for the service template or CSAR"
+    )
+    parser.add_argument(
+        "--instance-path", "-p",
+        help=".opera storage folder location"
+    )
+    parser.add_argument(
+        "--inputs", "-i", type=argparse.FileType("r"),
+        help="YAML file with inputs",
+    )
+    parser.add_argument("csar",
+                        type=argparse.FileType("r"),
+                        help="Cloud service archive or service template file")
+    parser.set_defaults(func=init)
+
+
+def init(args):
+    if args.instance_path and not path.isdir(args.instance_path):
+        raise argparse.ArgumentTypeError("Directory {0} is not a valid path!".format(args.instance_path))
+
+    storage = Storage(Path(args.instance_path) / ".opera") if args.instance_path else Storage(Path(".opera"))
+
+    if is_zipfile(args.csar.name):
+        initialize_compressed_csar(args, storage)
+    else:
+        storage.write(args.csar.name, "root_file")
+        initialize_service_template(args, storage)
+
+
+def initialize_service_template(args, storage):
+    try:
+        inputs = yaml.safe_load(args.inputs) if args.inputs else {}
+        storage.write_json(inputs, "inputs")
+    except Exception as e:
+        print("Invalid inputs: {}".format(e))
+        return 1
+
+    try:
+        ast = tosca.load(Path.cwd(), PurePath(args.csar.name))
+        template = ast.get_template(inputs)
+        template.instantiate(storage)
+    except ParseError as e:
+        print("{}: {}".format(e.loc, e))
+        return 1
+    except DataError as e:
+        print(str(e))
+        return 1
+
+    print("Service template was initialized")
+    return 0
+
+
+def initialize_compressed_csar(args, storage):
+    csars_dir = Path(storage.path) / "csars"
+    csars_dir.mkdir(exist_ok=True)
+    csar_name = args.csar.name
+
+    try:
+        inputs = yaml.safe_load(args.inputs) if args.inputs else {}
+        storage.write_json(inputs, "inputs")
+    except Exception as e:
+        print("Invalid inputs: {}".format(e))
+        return 1
+
+    try:
+        # validate csar
+        csar = CloudServiceArchive(csar_name, csars_dir)
+        tosca_service_template = csar.validate_csar()
+
+        # unzip csar and save the path to storage
+        csar_dir = csars_dir / Path("csar")
+        ZipFile(csar_name, 'r').extractall(csar_dir)
+        csar_tosca_service_template_path = csar_dir / tosca_service_template
+        storage.write(str(csar_tosca_service_template_path), "root_file")
+
+        # try to initiate service template from csar
+        ast = tosca.load(Path.cwd(), csar_tosca_service_template_path)
+        template = ast.get_template(inputs)
+        template.instantiate(storage)
+    except ParseError as e:
+        print("{}: {}".format(e.loc, e))
+        return 1
+    except DataError as e:
+        print(str(e))
+        return 1
+    except Exception as e:
+        print("Invalid CSAR: {}".format(e))
+        return 1
+
+    print("CSAR was initialized")
+    return 0
--- a/src/opera/commands/init.py
+++ b/src/opera/commands/init.py
@@ -34,7 +34,7 @@ def init(args):
     if args.instance_path and not path.isdir(args.instance_path):
         raise argparse.ArgumentTypeError("Directory {0} is not a valid path!".format(args.instance_path))
 
-    storage = Storage(Path(args.instance_path) / ".opera") if args.instance_path else Storage(Path(".opera"))
+    storage = Storage(Path(args.instance_path)) if args.instance_path else Storage(Path(".opera"))
 
     if is_zipfile(args.csar.name):
         initialize_compressed_csar(args, storage)
--- a/src/opera/commands/outputs.py
+++ b/src/opera/commands/outputs.py
@@ -30,7 +30,7 @@ def outputs(args):
     if args.instance_path and not path.isdir(args.instance_path):
         raise argparse.ArgumentTypeError("Directory {0} is not a valid path!".format(args.instance_path))
 
-    storage = Storage(Path(args.instance_path).joinpath(".opera")) if args.instance_path else Storage(Path(".opera"))
+    storage = Storage(Path(args.instance_path) / ".opera") if args.instance_path else Storage(Path(".opera"))
     root = storage.read("root_file")
     inputs = storage.read_json("inputs")
 
--- a/src/opera/commands/outputs.py
+++ b/src/opera/commands/outputs.py
@@ -30,7 +30,7 @@ def outputs(args):
     if args.instance_path and not path.isdir(args.instance_path):
         raise argparse.ArgumentTypeError("Directory {0} is not a valid path!".format(args.instance_path))
 
-    storage = Storage(Path(args.instance_path) / ".opera") if args.instance_path else Storage(Path(".opera"))
+    storage = Storage(Path(args.instance_path)) if args.instance_path else Storage(Path(".opera"))
     root = storage.read("root_file")
     inputs = storage.read_json("inputs")
 
--- a/src/opera/commands/undeploy.py
+++ b/src/opera/commands/undeploy.py
@@ -29,7 +29,7 @@ def undeploy(args):
     if args.instance_path and not path.isdir(args.instance_path):
         raise argparse.ArgumentTypeError("Directory {0} is not a valid path!".format(args.instance_path))
 
-    storage = Storage(Path(args.instance_path).joinpath(".opera")) if args.instance_path else Storage(Path(".opera"))
+    storage = Storage(Path(args.instance_path) / ".opera") if args.instance_path else Storage(Path(".opera"))
     root = storage.read("root_file")
     inputs = storage.read_json("inputs")
 
--- a/src/opera/commands/undeploy.py
+++ b/src/opera/commands/undeploy.py
@@ -29,7 +29,7 @@ def undeploy(args):
     if args.instance_path and not path.isdir(args.instance_path):
         raise argparse.ArgumentTypeError("Directory {0} is not a valid path!".format(args.instance_path))
 
-    storage = Storage(Path(args.instance_path) / ".opera") if args.instance_path else Storage(Path(".opera"))
+    storage = Storage(Path(args.instance_path)) if args.instance_path else Storage(Path(".opera"))
     root = storage.read("root_file")
     inputs = storage.read_json("inputs")
 
--- a/src/opera/parser/tosca/csar.py
+++ b/src/opera/parser/tosca/csar.py
@@ -1,0 +1,100 @@
+from pathlib import Path
+from tempfile import TemporaryDirectory
+from zipfile import ZipFile
+
+import yaml
+from opera.error import ParseError
+
+
+class CloudServiceArchive:
+    def __init__(self, csar_name, csar_folder_path):
+        self._csar_name = csar_name
+        self._csar_folder_path = csar_folder_path
+        self._tosca_meta = None
+        self._root_yaml_template = None
+        self._metadata = None
+
+    def validate_csar(self):
+        with TemporaryDirectory(dir=self._csar_folder_path) as tempdir:
+            ZipFile(self._csar_name, 'r').extractall(tempdir)
+
+            try:
+                meta_file_path = Path(tempdir) / "TOSCA-Metadata" / "TOSCA.meta"
+                if meta_file_path.exists():
+                    with meta_file_path.open() as meta_file:
+                        self._tosca_meta = yaml.safe_load(meta_file)
+
+                    self._validate_csar_version()
+                    self._validate_tosca_meta_file_version()
+                    template_entry = self._get_entry_definitions()
+                    self._get_created_by()
+                    self._get_other_definitions()
+
+                    # check if 'Entry-Definitions' points to an existing template file in the CSAR
+                    if not Path(tempdir).joinpath(template_entry).exists():
+                        raise ParseError('The file "{}" defined within "Entry-Definitions" in '
+                                         '"TOSCA-Metadata/TOSCA.meta" does not exist.'.format(template_entry), self)
+
+                    return template_entry
+                else:
+                    root_yaml_files = []
+
+                    root_yaml_files.extend(Path(tempdir).glob('*.yaml'))
+                    root_yaml_files.extend(Path(tempdir).glob('*.yml'))
+
+                    if len(root_yaml_files) != 1:
+                        raise ParseError("There should be one root level yaml "
+                                         "file in the root of the CSAR: {}.".format(root_yaml_files), self)
+
+                    with Path(root_yaml_files[0]).open() as root_template:
+                        root_yaml_template = yaml.safe_load(root_template)
+                        self._metadata = root_yaml_template.get('metadata')
+
+                    self._get_template_version()
+                    self._get_template_name()
+                    self._get_author()
+
+                    return root_yaml_files[0].name
+            except Exception as e:
+                raise ParseError("Invalid CSAR structure: {}".format(e), self)
+
+    def _validate_csar_version(self):
+        csar_version = self._tosca_meta.get('CSAR-Version')
+        if csar_version and csar_version != 1.1:
+            raise ParseError('CSAR-Version entry in the CSAR {} is '
+                             'required to denote version 1.1".'.format(self._csar_name), self)
+        return csar_version
+
+    def _validate_tosca_meta_file_version(self):
+        tosca_meta_file_version = self._tosca_meta.get('TOSCA-Meta-File-Version')
+        if tosca_meta_file_version and tosca_meta_file_version != 1.1:
+            raise ParseError('TOSCA-Meta-File-Version entry in the CSAR {} is '
+                             'required to denote version 1.1".'.format(self._csar_name), self)
+        return tosca_meta_file_version
+
+    def _get_entry_definitions(self):
+        if 'Entry-Definitions' not in self._tosca_meta:
+            raise ParseError('The CSAR "{}" is missing the required metadata "Entry-Definitions" in '
+                             '"TOSCA-Metadata/TOSCA.meta".'.format(self._csar_name), self)
+        return self._tosca_meta.get('Entry-Definitions')
+
+    def _get_created_by(self):
+        return self._tosca_meta.get('Created-By')
+
+    def _get_other_definitions(self):
+        return self._tosca_meta.get('Other-Definitions')
+
+    def _get_template_version(self):
+        if "template_version" not in self._metadata:
+            raise Exception('The CSAR "{}" is missing the required '
+                            'template_version in metadata".'.format(self._csar_name))
+        return self._metadata.get('template_version')
+
+    def _get_template_name(self):
+        if "template_version" not in self._metadata:
+            raise ParseError('The CSAR "{}" is missing the required'
+                             ' template_name in metadata".'.format(self._csar_name), self)
+        return self._metadata.get('template_name')
+
+    def _get_author(self):
+        return self._metadata.get('template_author')
--- a/src/opera/commands/__init__.py
+++ b/src/opera/commands/__init__.py
@@ -2,5 +2,6 @@ from opera.commands import (
     outputs,
     deploy,
     undeploy,
-    validate
+    validate,
+    init
 )
--- a/src/opera/storage.py
+++ b/src/opera/storage.py
@@ -22,3 +22,7 @@ class Storage:
 
     def read_json(self, *path):
         return json.loads(self.read(*path))
+
+    def exists(self, *path):
+        return (self.path / pathlib.PurePath(*path)).exists()
+
--- a/tests/integration/misc-tosca-types/TOSCA-Metadata/TOSCA.meta
+++ b/tests/integration/misc-tosca-types/TOSCA-Metadata/TOSCA.meta
@@ -1,0 +1,4 @@
+TOSCA-Meta-File-Version: 1.1
+CSAR-Version: 1.1
+Created-By: OASIS TOSCA TC
+Entry-Definitions: service-template.yaml
--- a/tests/integration/misc-tosca-types/runme.sh
+++ b/tests/integration/misc-tosca-types/runme.sh
@@ -4,8 +4,16 @@ set -euo pipefail
 # get opera executable
 opera_executable="$1"
 
-# perform integration test
+# perform integration test with service template
 $opera_executable validate -i inputs.yaml service-template.yaml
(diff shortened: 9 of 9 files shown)


matejart:
Thank you for this work, @anzoman . However, I do not think that this is the right approach to supporting the comressed CSAR. I will once again point to how I view the CSAR should be used: #39 

My fundamental issue with this proposal is the requirement to unpack again a CSAR - the same one or a completely different one. The `undeploy` needs to work with the same service template that was used in the deploy, so it should be the orchestrator's responsibility to store it when it enters the deployment lifecycle, and use the same one throught the lifecycle until the very end. So using the temporary directory mechanism which is gone at the end of `deploy` isn't suitable, because those files need to be around until the end of `undeploy`.

In other words, and as Tadej's the comment that got deleted in this PR also says, I maintain that the unpacking of the CSAR needs to happen in the `init` command and into an empty `$PWD`. If going the `init` path right now is too disruptive, I would settle for the midpoint where `deploy` does that, but again in an empty `$PWD`. If the user uses `--instance-path`, then that one would be used instead of `$PWD`, but, again, I think it should be empty.

anzoman:
@matejart thank you for your comments. I see the intention to use `opera package` and `opera init` commands before deploying the compressed CSAR file but I was thinking of a simple deployment of the zipped CSAR invoked just by calling `opera deploy` and therefore I opened this PR. And I completely agree that my changes can be problematic and that the (un)zipped file should be stored rather that linked multiple times. So in this case do you suggest that I close this PR or could these changes be useful for something?

src/opera/commands/deploy.py
```diff
@@ -57,3 +73,30 @@ def deploy(args):
...
 
     return 0
+
+
+def deploy_compressed_csar(args, storage):
+    with TemporaryDirectory(dir=storage.path) as tempdir:
+        ZipFile(args.csar.name, 'r').extractall(tempdir)
+
+        try:
+            inputs_file = open(tempdir + "/" + args.inputs, 'r')
```
matejart, line 83: Inputs shouldn't be packaged into the CSAR. They are a separate file.

src/opera/commands/deploy.py
```diff
@@ -57,3 +73,30 @@ def deploy(args):
...
+            inputs = yaml.safe_load(inputs_file) if args.inputs else {}
+            storage.write_json(inputs, "inputs")
+        except Exception as e:
+            print("Invalid inputs: {}".format(e))
+            return 1
+
+        try:
+            ast = tosca.load(Path.cwd(), PurePath(tempdir + "/" + args.template))
```
matejart, line 91: Since we're using pathlib, let's use it all the way and not mix string concatenation.

matejart:
One more note: the orchestrator definitely needs to sanitize the zip inputs to prevent any [path traversal vulnerabilities](https://www.cisecurity.org/advisory/a-vulnerability-in-zoom-client-could-allow-for-arbitrary-code-execution_2020-076/) when accepting user-supplied artifacts.

src/opera/commands/deploy.py
```diff
@@ -19,12 +20,16 @@ def add_parser(subparsers):
...
     )
     parser.add_argument(
-        "--inputs", "-i", type=argparse.FileType("r"),
+        "--inputs", "-i", type=str,
         help="YAML file with inputs",
     )
+    parser.add_argument(
+        "--template", "-t", type=str,
+        help="TOSCA YAML service template file",
+    )
```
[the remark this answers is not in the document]
anzoman: Ok, so now when the CSAR is being initialized I look for the template in `Entry-Definitions` and use it later when deploy is called.

src/opera/commands/deploy.py
anzoman: This should be fixed now.

src/opera/commands/deploy.py
anzoman: Ok, so now the inputs are left out of CSARs and provided optionally as a separate file.

anzoman:
Thanks @matejart for all the comments and suggestions. I implemented the first version of `opera init` command and was trying to stick to the [TOSCA YAML v1.3 CSAR section](https://docs.oasis-open.org/tosca/TOSCA-Simple-Profile-YAML/v1.3/cos01/TOSCA-Simple-Profile-YAML-v1.3-cos01.html#_Toc26969474) in order to validate the compressed CSAR files. I have also modified the `opera deploy` command as you proposed and refactored integration tests so that the new commands are now also tested. But I'm not completely sure that I prevented all the possible zip slip risks so you can probably give me some advice about this here.

dradX:
@matejart , @anzoman 

Just a question regarding `opera init`. From the point of view of the orchestrator the inputs file is used to pass parameters to the execution of the `deploy`/`undeploy` standard workflows. The inputs can be used for passing sensitive information to the execution of `deploy` and `undeploy` - [TOSCA](https://docs.oasis-open.org/tosca/TOSCA-Simple-Profile-YAML/v1.3/csprd01/TOSCA-Simple-Profile-YAML-v1.3-csprd01.html#_Toc9262284) has quite a few examples of passing passwords to the execution of the service template. Is persisting this information  in the orchestrators internal environment besides the blueprint itself a good idea?
Shouldn't it be more prudent to consume the `inputs` by the orchestrator at `deploy` runtime execution, thus minimizing any risk of persisting sensitive data anywhere in the `opera` environment?

anzoman:
@dradX this is a good point because if you have your passwords within the key-value `inputs.yaml` you probably don't want them to be stored any longer than just for the execution or just when they are addessed by `get_input`. I think that the security might also be the main reason why `inputs.yaml` are not directly part of the CSAR. So maybe it is better to keep inputs away from the init and to use the them only with deploy option. But on the other hand this would mean that without the inputs (if they are being used) init command could not fully initialize the deployment. What do you think @matejart?

matejart:
@anzoman  The reason for not putting inputs into the CSAR is because the CSAR is reusable, static, immutable archive, while inputs are changeable and deploy-specific.

@dradX @anzoman in my opinion, we need to treat the whole opera store as sensitive. It is not just passwords in the inputs that are sensitive, the deployment will probably generate new unique identities, keys and tokens, which could also be used to gain access to parts of the deployment that should be restricted. So focusing on inputs only would only address one part of the issue.
The other aspect is that the orchestrator needs the secrets to be available in plaintext, which means that they should be there in plaintext or be decrypted just for a the time when they get passed to the runner.

So I would consider going into the direction where the instance model and everything in the opera store gets encrypted with a key or password, for which the tool prompts before running init, deploy, undeploy, etc. Or, it would take a password file from the hard drive. Either way, I would consider this an enterprise-grade, paid-for extension to the regular opera. Until then, we should set the permissions to the `.opera` folder to be only available to the user.

dradX:
@matejart We can't possibly know what the user has to do in the blueprint - created access tokens etc. 
But we should be able to address in the right way (with the appropriate workflow) the issues that we _can address_ and inputs is one that can be easily addressed in this way - by providing inputs on deploy.

cankarm:
Hi lads,

First, thumbs up that CSAR zip has everything, service template + inputs, BUT inputs are not necessary already filled with all sensitive data. It needs to be just a file with the appropriate format. All details which files and what names required for a start are in "meta file" - https://docs.oasis-open.org/tosca/TOSCA-Simple-Profile-YAML/v1.3/csprd01/TOSCA-Simple-Profile-YAML-v1.3-csprd01.html#_Toc9262338. I see that you already found that and this is also the way how we get the CSAR form some tools, e.g. Winery.

At the "unpack" and "init" or whatever we choose, the CSAR is prepared for deploy. So we internally store the content that needs to be deployed. User should be able to change the inputs at that point or if very sensitive, might be able to add it at deploy? At this time a final snapshot of the deploy needs to be stored (probably with unique time id) so we have a version saved for undeploy.

@dradX : Just for another aspect:
Question1: We may allow to execute deploy with "empty or unconfigure inputs.yaml" and prompt user for every undefined variable.
Question2: There are deployments that probably don't need inputs, and don't require an init at all. So "unpack" and "deploy" is enough. 


Maybe we will need a developers summit as we have too different use-cases in front of us to simply understand them and provide a solution to rule them all. :D

src/opera/parser/tosca/csar.py
```diff
@@ -0,0 +1,101 @@
...
+
+                    self._get_template_version()
+                    self._get_template_name()
+                    self._get_author()
+
+                    return root_yaml_files[0].name
+            except Exception as e:
+                raise ParseError("Invalid CSAR structure: {}".format(e), self)
+
+    def _get_csar_version(self):
```
matejart, line 62:
```suggestion
    def _validate_csar_version(self):
```

I suggest renaming these methods, because they act more a validator and less a getter.

src/opera/parser/tosca/csar.py
```diff
@@ -0,0 +1,101 @@
...
+        csar_version = self._tosca_meta.get('CSAR-Version')
+        if csar_version and csar_version != 1.1:
+            raise ParseError('CSAR-Version entry in the CSAR {} is '
+                             'required to denote version 1.1".'.format(self._csar_name), self)
+        return csar_version
+
+    def _get_tosca_meta_file_version(self):
```
matejart, line 69:
```suggestion
    def _validate_tosca_meta_file_version(self):
```

src/opera/parser/tosca/csar.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def validate_csar(self):
+        with TemporaryDirectory(dir=self._csar_folder_path) as tempdir:
+            ZipFile(self._csar_name, 'r').extractall(tempdir)
+
+            try:
+                if Path(tempdir).joinpath("TOSCA-Metadata").joinpath("TOSCA.meta").exists():
+                    meta_file = open(Path(tempdir).joinpath("TOSCA-Metadata").joinpath("TOSCA.meta"), 'r')
+                    self._tosca_meta = yaml.safe_load(meta_file)
+                    meta_file.close()
```
matejart, line 25:
```suggestion
                meta_file_path = Path(tempdir) / "TOSCA-Metadata" / "TOSCA.meta"
                if meta_file_path.exists():
                    with meta_file_path.open() as meta_file:
                        self._tosca_meta = yaml.safe_load(meta_file)
```

matejart: Shortens the lines, makes the pathlib usage consistent accross the project, and uses a safe way to handle files.

src/opera/parser/tosca/csar.py
```diff
@@ -0,0 +1,101 @@
...
+                    root_yaml_files.extend(Path(tempdir).glob('*.yml'))
+
+                    if len(root_yaml_files) != 1:
+                        raise ParseError("There should be one root level yaml "
+                                         "file in the root of the CSAR: {}.".format(root_yaml_files), self)
+
+                    root_template = open(Path(root_yaml_files[0]), "r")
+                    root_yaml_template = yaml.safe_load(root_template)
+                    self._metadata = root_yaml_template.get('metadata')
+                    root_template.close()
```
matejart, line 52:
```suggestion
                    with Path(root_yaml_files[0]).open() as root_template:
                        root_yaml_template = yaml.safe_load(root_template)
                        self._metadata = root_yaml_template.get('metadata')
```

anzoman:
@matejart thank you for the explanation. I agree that sensitive information can be everywhere, even in the CSAR itself and I really like your enterprise idea of encrypting opera storage with passwords.

And regarding the inputs I found one interesting thing (looking at  TOSCA YAML v1.3 13.2.7 section titled Pass Inputs and Retrieve Results or Errors):

> Orchestrators must pass any required inputs to the artifact processor. Some processors could take inputs through environment variables, but others may prefer command line arguments. Named or positional command line arguments could be used. TOSCA must be very specific about the mechanism for passing input data to processors for each type of artifact.
> 
> Similarly, artifact processors must also pass results from operations back to orchestrators so that results values can be reflected as appropriate in node properties and attributes. If the operation fails, error codes may need to be returned as well. TOSCA must be very specific about the mechanism for returning results and error codes for each type of artifact. 

So as far as I understand `inputs.yaml` is just one way to provide inputs to the opera orchestartor. We can have in mind that it is possible to pick the other ways (env vars, CLI args) if needed.

@cankarm thumbsup for the proposed summit :smiley:

dradX:
@cankarm 
Setting inputs file is the standard way configure the execution of blueprint deployment - hopefully we can agree on that. Using the _inputs_ at `deploy`/`undeploy` supports the workflow without the need to save the inputs with the CSAR files - saving _inputs_ along with the CSAR is definitely not a must for deploying blueprints and should be avoided. 
And I agree we should have developers summit.

matejart:
@dradX I understand your position, but that's just the thing: we _don't_ really control the inputs once they become an integral part of the instance model, and that is excatly what they are: their values get copied into properties and attributes and stored on disk verbatim or in some derived form. So requiring to supply the inputs time with each call doesn't help in keeping things secret, while it introduces possible unexpected side effects where an input used in one way for deploy is different from the one needed in the undeploy.

So this is the answer to both questions: input values become a part of the instance managed by the orchestrator, so we give it at the entry point, which now is `init`. Once we support changing a deploy, I imagine that you'll be able to provide a different set of inputs, at which point the orchestrator will have an opportunity to obtain differences due to the different inputs and act accordingly.
Considering that at the phase of init, currently nothing actually happens with the inputs value except them being saved with the instance, we kept the optional `--inputs` switch for the `deploy` for now. This way, you can provide them at init or deploy, or both. 

> Why would the user have to call init(blueprint/inputs) and then deploy without inputs?

They woldn't deploy without inputs, the deployment uses the inputs supplied in init.

dradX:
@matejart 
Maybe I sense a slight misunderstanding when we are writing about the instance:

> .. inputs once they become an integral part of the instance model, and that is excatly what they are: their values get copied into properties and attributes and stored on disk verbatim or in some derived form.

An **instance** of the deployment is created exactly in the process of `opera deploy` and not before or after that. I was sure that `opera deploy` is when we start the deployment and create an execution tree from the blueprint and take the inputs for the nodes that need them - and this is not the description of the `opera init˛` or perhaps I missed something?

matejart:
@cankarm thank you for pitching in with your perspectives.

> First, thumbs up that CSAR zip has everything, service template + inputs, BUT inputs are not necessary already filled with all sensitive data. It needs to be just a file with the appropriate format. 

I view CSAR in the same way as a JAR file or a Linux elf or Windows EXE file: a pre-built artifact that is aimed to be consumed by the executor. The user could unpack it, but shouldn't have to. So placing anything inside the CSAR that needs to be retrieved by the user shouldn't be necessary. Of course, the users are free to supply templated or abbreviated inputs files on top of a CSAR if they like to.

> All details which files and what names required for a start are in "meta file" - https://docs.oasis-open.org/tosca/TOSCA-Simple-Profile-YAML/v1.3/csprd01/TOSCA-Simple-Profile-YAML-v1.3-csprd01.html#_Toc9262338. I see that you already found that and this is also the way how we get the CSAR form some tools, e.g. Winery.

Yes, @anzoman has implemented this already.

> At this time a final snapshot of the deploy needs to be stored (probably with unique time id) so we have a version saved for undeploy.

There is really always just one snapshot of the instance, no versioning. The orchestrator transitions between the snapshots when a workflow is invoked.

> Question1: We may allow to execute deploy with "empty or unconfigure inputs.yaml" and prompt user for every undefined variable.

I don't think the orchestrator shoud do that, but you could have a wrapper around it to do that if needed. Still, having to sit at the console to answer questions doesn't sound to me like a useful way for automating things.

matejart:
> @matejart
> An **instance** of the deployment is created exactly in the process of `opera deploy` and not before or after that. I was sure that `opera deploy` is when we start the deployment and create an execution tree from the blueprint and take the inputs for the nodes that need them - and this is not the description of the `opera init˛` or perhaps I missed something?

Yes, you are correct about the **instance**. What I am referring to, is the instance **model**, which is the orchestrator's internal representation about what state the instance at some point in time is. Right now, this comes to be in the `deploy` phase, like you observed yourself, so it's fine to supply inputs there. But I can imagine that the `init` would just as well be able to build one, just with all the nodes not being initialized yet. Just some future thinking.

matejart:
@anzoman thank you for pointing out the section 13, but it looks like a bunch of half-cooked and non-normative ideas. But sure, for good and valid use cases, other means of supplying inputs may be worth considering and supporting. But the dynamics of ingesting them should remain the same as now where it all comes from a YAML file.

cankarm:
@matejart, @dradX 
You definitely need to talk in person ;)

> >    Question1: We may allow to execute deploy with "empty or unconfigure inputs.yaml" and prompt user for every undefined variable.
> 
> I don't think the orchestrator should do that, but you could have a wrapper around it to do that if needed. Still, having to sit at the console to answer questions doesn't sound to me like a useful way for automating things.

Sure, I agree. This is not elegant, but if you are very very paranoid that you do not want to store anything anywhere this a dirty hack. So the credentials live only in memory and they never get stored. Even retrieving them from ENV vars is not easily possible.

For me even unpack and init is something that I would like to avoid when automating (not testing or developing, but just production deploy). But frequently I try to see this whole conversation through the eyes of Orchestrator as a Service, where I like to have multiple steps that can be bundled in calls towards the API. But there most of this problems can be solved differently, I guess.

dradX:
@matejart 
I understand the so the **instance model** can of course be built with presenting the CSAR files and without any inputs - this is `opera init` no need for inputs there to initialize. 
While the instance where the instance model gets instantiated is created is in `opera deploy` - this is where to user commits to use a set of inputs for creating an **instance** from the instance model - so here we do _have to have inputs_. 
So if the user wanted to set up **another instance** why not call the `opera deploy` with inputs when needed instead of the need to call another opera init -> opera deploy?
It would be good to know if there is a plan to change this at some point.

cankarm:
> @cankarm
> Setting inputs file is the standard way confto igure the execution of blueprint deployment - hopefully we can agree on that. Using the inputs at deploy/undeploy supports the workflow without the need to save the inputs with the CSAR files - saving inputs along with the CSAR is definitely not a must for deploying blueprints and should be avoided.
> And I agree we should have developers summit.

@dradX: I thought that I was clear, sorry. Similar as Matej explained CSAR is a package. The exact values of inputs _should_ be provided along. But a template how you should form and serve the inputs should be in CSAR. So a template-inputs file should be in CSAR and should be referenced in service.meta. So I guess that it is a must to have _template of inputs_ in CSAR. Otherwise you always need two (CSAR + inputs template) files to form a template of application. So another ZIP that packs CSAR + template_input.yaml.

But again CSAR is a blueprint. So this does not imply that exact values of your deployment must be in inputs file packet in your CSAR. Only template of it.

dradX:
@cankarm  
> 
> @dradX: I thought that I was clear, sorry. Similar as Matej explained CSAR is a package. The exact values of inputs _should_ be provided along. But a template how you should form and serve the inputs should be in CSAR. So a template-inputs file should be in CSAR and should be referenced in service.meta. So I guess that it is a must to have _template of inputs_ in CSAR. Otherwise you always need two (CSAR + inputs template) files to form a template of application. So another ZIP that packs CSAR + template_input.yaml.
> 
> But again CSAR is a blueprint. So this does not imply that exact values of your deployment must be in inputs file packet in your CSAR. Only template of it.

This is a completely new use case I guess -  an _input template_ is an unknown artifact and by the standard and if it is used inside the CSAR then its not _inputs_. By _service.meta_ you mean Tosca.meta?

matejart:
@cankarm 

> Sure, I agree. This is not elegant, but if you are very very paranoid that you do not want to store anything anywhere this a dirty hack. So the credentials live only in memory and they never get stored. 

An instance model gets stored on a hard drive, and an instance model contains copied or digested inputs values. If paranoid, you need to encrypt everything on the hard drive.

> But frequently I try to see this whole conversation through the eyes of Orchestrator as a Service, where I like to have multiple steps that can be bundled in calls towards the API. 

`\w+` as a Service means that your orchestrator lives on someone else's computer, so you need to entrust everything from CSAR to the passwords in the inputs to that computer. There are of course ways to point-to-point-encrypt everything and then limit access to the files (which themselves may live in an encrypted envelope). But just-in-time prompting for secrets makes even less sense in this use case.

> So a template-inputs file should be in CSAR and should be referenced in service.meta. So I guess that it is a must to have template of inputs in CSAR. 

The `topology_template.inputs` is the inputs template that is a part of the service definition. It is not a separate file, or at least it doesn't have to be, and doesn't get its own meta file reference in CSAR. But we can implement `opera inputs` command which parses the service template and dumps the inputs definition. Cloudify has that, if the memory serves correctly.

@dradX 
> So if the user wanted to set up another instance why not call the opera deploy with inputs when needed instead of the need to call another opera init -> opera deploy?

opera currently supports only one deployed instance from one instance folder (such as `$PWD/.opera`). If you want to make a new instance, you need to undeploy (or abandon) the previous one first. But if you need multiple instances in parallel, they will each have to live in their separate instance folder, each being init-ed in turn. This is to make things simpler and cleaner in terms of everything that needs to be in place for (un)deployments.

cankarm:
My bad.
@dradX  yes I mean tosca.meta.

@matejart a command that generates inputs template file is OK. But if orchestrator does not support this generation of an input file then the user has more work to generate this file for himself. I understand that this is not by the standard, but I would not object to the solution to have another file already generated in CSAR.

src/opera/commands/deploy.py
```diff
@@ -22,9 +22,8 @@ def add_parser(subparsers):
         "--inputs", "-i", type=argparse.FileType("r"),
         help="YAML file with inputs",
```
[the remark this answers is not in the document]
anzoman: This was refactored as you suggested.

src/opera/commands/init.py
```diff
@@ -0,0 +1,107 @@
...
+                        type=argparse.FileType("r"),
+                        help="Cloud service archive or service template file")
+    parser.set_defaults(func=init)
+
+
+def init(args):
+    if args.instance_path and not path.isdir(args.instance_path):
+        raise argparse.ArgumentTypeError("Directory {0} is not a valid path!".format(args.instance_path))
+
+    storage = Storage(Path(args.instance_path).joinpath(".opera")) if args.instance_path else Storage(Path(".opera"))
```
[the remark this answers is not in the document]
anzoman: Thanks for the advice, I've shortened the paths. 

And regarding the `.opera` appending before we implemented the instance path option opera had `storage = Storage(Path(".opera"))` so this was kept if `--instance-path` option would not be supplied.  But if this option is used do you suggest that we just point the opera to that directory without creating the `.opera` there? Or do you have something else in mind?

src/opera/commands/init.py
```diff
@@ -0,0 +1,107 @@
...
+    except DataError as e:
+        print(str(e))
+        return 1
+
+    print("Service template was initialized!")
+    return 0
+
+
+def initialize_compressed_csar(args, storage):
+    csars_dir = Path(storage.path).joinpath("csars")
```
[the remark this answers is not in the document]
anzoman: Done.

src/opera/commands/init.py
```diff
@@ -0,0 +1,107 @@
...
+        print("Invalid inputs: {}".format(e))
+        return 1
+
+    try:
+        # validate csar
+        csar = CloudServiceArchive(csar_name, csars_dir)
+        tosca_service_template = csar.validate_csar()
+
+        # unzip csar and save the path to storage
+        csar_dir = csars_dir.joinpath(Path(csar_name).stem)
```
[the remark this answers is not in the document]
anzoman: Refactored.

src/opera/commands/init.py
```diff
@@ -0,0 +1,107 @@
...
+        ZipFile(csar_name, 'r').extractall(csar_dir)
+        csar_tosca_service_template_path = csar_dir.joinpath(tosca_service_template)
```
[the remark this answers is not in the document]
anzoman: This change was also made.

src/opera/commands/init.py
```diff
@@ -0,0 +1,107 @@
...
+        print("{}: {}".format(e.loc, e))
+        return 1
+    except DataError as e:
+        print(str(e))
+        return 1
+    except Exception as e:
+        print("Invalid CSAR: {}".format(e))
+        return 1
+
+    print("CSAR was initialized!")
```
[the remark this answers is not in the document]
anzoman: The exclamation mark has been erased.

src/opera/commands/undeploy.py
```diff
@@ -16,6 +18,9 @@ def add_parser(subparsers):
         "--instance-path", "-p",
         help=".opera storage folder location"
     )
+    parser.add_argument("-compressed-csar", "-c",
+                        type=argparse.FileType("r"),
+                        help="Compressed cloud service archive file")
```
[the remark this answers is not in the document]
anzoman: Yes, true I have forgotten about this. I have reversed this to the state before initial commits.

src/opera/parser/tosca/csar.py
anzoman: Ok, I agree and this was modified as you suggested.

src/opera/parser/tosca/csar.py
anzoman: Done.

src/opera/parser/tosca/csar.py
anzoman: Good point, I've added your suggested changes.

src/opera/parser/tosca/csar.py
anzoman: These changes were also included.

anzoman:
@matejart I've refactored the code and now most of your suggested changes should be included. And you asked about `opera undeploy` - yes this command is still in its initial state without any main changes. So you also propose changes for this if there should be any. Now when the CSAR is unpacked it it stored in `.opera/csars` directory within the folder that has the same name as the CSAR (without the `.csar` or `zip` extension) - do you think that we will need to use UUIDs for the name of the these CSAR opera storage folders or is it okay if we keep the CSAR's name as the name of the opera csar storage folder?

We also have to decide what to do with validate command. I see that it doesn't have `--instance-path` option which will probably need to be included and also we will have to make some changes if we want to use validate with the zipped CSAR files - but still - now `opera init` actually also includes the same thing that is present in validate - with that I mean getting/parsing/loading the TOSCA service template so both commands include the code below:

```python
ast = tosca.load(Path.cwd(), PurePath(args.csar.name))
ast.get_template(inputs)
```

src/opera/commands/deploy.py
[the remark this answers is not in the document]
anzoman: But instead of the `csar` positional argument I used `template` to mark that deploy is used with TOSCA service templates and not directly with CSAR files.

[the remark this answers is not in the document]
anzoman: This `template` positional argument is now optional in case that user wants to do `opera deploy template.yaml` as before but I can remove this argument if you think that it is necessary just to kep the inputs argument and to move everything else to init.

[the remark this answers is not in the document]
matejart: I'm good with being able to provide path to a csar or, like in the past, a path to a yaml.

matejart:
@anzoman the name of the folder containing an unpacked csar is up to us. I see an issue when the input might be insance.csar, clashing with the other folders in .opera. So I'd just name that folder csar and, maybe, save the name in the store. There is always just one csar per instance. So the undeploy can use the yaml used in the deploy.

matejart:
As for the validate, I see the only real use case as the one where we are developing a service template and use it with a specific yaml, which is how it works now. For a csar, I'd say that the expectance is that it is valid, so I don't see a lot of value in supporting in that way. Particularly since, as you say, init does that. So in this respect, it should be fine the way it is now.

src/opera/commands/deploy.py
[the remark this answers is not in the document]
anzoman: Ok great :+1:

anzoman:
@matejart thanks for your suggestions!

So now the name of the folder containing an unpacked csar is set to `csar` as  there is always just one csar per instance.

And regarding the opera validate I kept it just like as it was before without any new changes.

anzoman:
@matejart , @dradX and @cankarm so to be clear after the new changes from this PR the opera CLI commands will be used like below (I've gathered some possible usages of the commands):

- **validate:**

```console
# validate service template
opera validate service-template.yaml

# validate service template (also with inputs)
opera validate --inputs inputs.yaml service-template.yaml
```



# initialize single  TOSCA service template
opera init service-template.yaml

# initialize the CSAR (from the prepared compressed CSAR file) 
# also provide inputs and relocate .opera to some other folder
opera init --inputs inputs.yaml --instance-path ./test-dir test.csar

# deploy the initialized CSAR/service template
opera deploy

# deploy the initialized CSAR/service template 
# also override inputs and use opera storage from some other folder 
opera deploy -i inputs.yaml -p test-folder/test-subfolder/

# deploy single service template (like before the PR)
opera deploy sevice-template-yaml

# deploy single service template (like before the PR)
# also use inputs
opera deploy -i inputs.yaml sevice-template-yaml

# take a look at opera outputs
opera outputs

# take a look at opera outputs
# provide path to opera storage folder where the outputs might be
opera outputs --instance-path ./test-dir

# undeploy a deployed instance
opera undeploy

# undeploy a deployed instance
# point to the opera storage where the CSAR/service template is stored
opera undeploy -p test-folder/test-subfolder/
```

File: src/opera/commands/init.py (hunk 9 above)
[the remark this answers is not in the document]
Comment (matejart) [reply src/opera/commands/init.py]: Yes, I think the `--instance-path` path can be used as is without creating the `.opera` subfolder. I worry that otherwise there might be some confusion as to why the target folders appear to be empty, since they would only contain the hidden subfolder.

File: src/opera/commands/deploy.py [hunk 15]
```diff
@@ -20,32 +20,48 @@ def add_parser(subparsers):
...
     parser.set_defaults(func=deploy)
 
 
 def deploy(args):
     if args.instance_path and not path.isdir(args.instance_path):
         raise argparse.ArgumentTypeError("Directory {0} is not a valid path!".format(args.instance_path))
 
-    storage = Storage(Path(args.instance_path).joinpath(".opera")) if args.instance_path else Storage(Path(".opera"))
-    storage.write(args.csar.name, "root_file")
+    storage = Storage(Path(args.instance_path) / ".opera") if args.instance_path else Storage(Path(".opera"))
```
matejart, line 36:
```suggestion
    storage = Storage(Path(args.instance_path)) if args.instance_path else Storage(Path(".opera"))
```
Suggestion as discussed in the main thread.

src/opera/commands/init.py
```diff
@@ -0,0 +1,107 @@
...
+                        type=argparse.FileType("r"),
+                        help="Cloud service archive or service template file")
+    parser.set_defaults(func=init)
+
+
+def init(args):
+    if args.instance_path and not path.isdir(args.instance_path):
+        raise argparse.ArgumentTypeError("Directory {0} is not a valid path!".format(args.instance_path))
+
+    storage = Storage(Path(args.instance_path) / ".opera") if args.instance_path else Storage(Path(".opera"))
```
matejart, line 37:
```suggestion
    storage = Storage(Path(args.instance_path)) if args.instance_path else Storage(Path(".opera"))
```
As discussed in the main thread.

src/opera/commands/outputs.py
```diff
@@ -30,7 +30,7 @@ def outputs(args):
     if args.instance_path and not path.isdir(args.instance_path):
         raise argparse.ArgumentTypeError("Directory {0} is not a valid path!".format(args.instance_path))
 
-    storage = Storage(Path(args.instance_path).joinpath(".opera")) if args.instance_path else Storage(Path(".opera"))
+    storage = Storage(Path(args.instance_path) / ".opera") if args.instance_path else Storage(Path(".opera"))
```
matejart, line 33:
```suggestion
    storage = Storage(Path(args.instance_path)) if args.instance_path else Storage(Path(".opera"))
```
Same as above.

src/opera/commands/undeploy.py
```diff
@@ -23,7 +23,7 @@ def undeploy(args):
     if args.instance_path and not path.isdir(args.instance_path):
         raise argparse.ArgumentTypeError("Directory {0} is not a valid path!".format(args.instance_path))
 
-    storage = Storage(Path(args.instance_path).joinpath(".opera")) if args.instance_path else Storage(Path(".opera"))
+    storage = Storage(Path(args.instance_path) / ".opera") if args.instance_path else Storage(Path(".opera"))
```
matejart, line 26:
```suggestion
    storage = Storage(Path(args.instance_path)) if args.instance_path else Storage(Path(".opera"))
```
As above.

src/opera/commands/init.py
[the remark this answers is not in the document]
anzoman: Ok, I've made changes that omit appending `.opera` by default that and have put this in a separate commit along with the explanation.

src/opera/commands/deploy.py
anzoman: Done.

src/opera/commands/init.py
anzoman: Resolved.

src/opera/commands/outputs.py
anzoman: Refactored.

src/opera/commands/undeploy.py
anzoman: This was also modified as you suggested.

anzoman:
@matejart automatic appending  of the .opera path (when using `--instance-path` option) has been thrown out. The new integration tests pass successfully and everything seemed fine when I was testing the changes locally. You can try to test this yourself if needed and then I guess that will be it.

matejart:
@anzoman there seem to be conflicts blocking the merge.

anzoman:
@matejart aha I see. But no problem, I will do a rebase on master branch.

anzoman:
@matejart now it should be possible to merge this. I've also tested my changes locally after the rebase and everything seemed fine as before.

Issue #47, opened 2020-05-29
Transfer opera documentation to this repository


anzoman:
## Description
This use case addresses the possibility to relocate xOpera's documentation. Currently the documentation for opera orchestrator is available here: https://xlab-si.github.io/xopera-docs/ and originates from this repository: https://github.com/xlab-si/xopera-docs. The documentation was initially created to present opera orchestration tool as a micosite for Horizon 2020 RADON project and for the fast creation it was put in a separate repository but there's no doubt that through time the documentation will be extended and will explain opera's components in detail. Therefore I think that we should keep things considering opera together and so that opera's docs could also be a part of this repository. 

## Prerequisites

- Read the Docs account (optional)

## Steps

1. I suggest that we first create `docs` folder here at the root of opera's repository. This directory would contain all the documentation files (.rst files) and other assets that would be displayed in our docs.

2. The next step would feature creating a new PR that would move the current documentation files here. For the documentation rendering we can use Sphinx tool and prepare a Makefile that builds the HTML documentation files which can be actually hosted somewhere. And as we are already here for the hosting we can use GitHub Pages.

3. So after the documentation source files are fully prepared we can create a special branch called `gh-pages`, build the docs and push them on this branch (here we shouldn't forget the `.nojekyll file` in order to prevent GitHub Pages from needlessly running Jekyll themes.). 

4. And eventually under the repository settings we can mark to build the xOpera's GitHub Pages site from `gh-pages` branch. 

## Expected results
To have the documentation source files in this repository (in `docs` folder) and the documentation available on: https://xlab-si.github.io/xopera-opera/.

## Note
There is also another way of doing all that which is to create an account at https://readthedocs.org/, upload the documentation there and connect this to GitHub but this is suitable only if we want to use Read the Docs style for our documentation.

sstanovnik:
Resolved by #60.

PR #60, opened 2020-06-16
Introduce the very first opera documentation

anzoman:
In this commit we are paving the way for establishing the initial
documentation for xOpera orchestrator. The documentation is currently
in the form of .rst files and is prepared to be build with Read the
Docs and Sphinx documentation tool generator. We prepared a Makefile to
be able to build our documentation locally.

With these changes we set the structure for hosting our docs on GitHub
Pages. To accomplish that we need to separate source and build files
into two branches. The source .rst documentation files will reside in
docs folder of the master branch whereas the rendered output HTML
files will be found on gh-pages branch which will represent the
publishing source for our GitHub Pages documentation site. To
constantly keep the xOpera's documentation up to date we created a
CircleCI worklflow that builds and pushes the generated documentation
files to gh-pages branch. And last but not least, we had to include is
a .nojekyll file in order to bypass Jekyll processing on GitHub Pages.

Our first documentation is not perfect and was initially meant to be a
tool introduction page or microsite for European Horizon 2020 RADON
project. This commit is also directly linked to the issue #47 that
suggested moving the documentation from the separate repository here.


.circleci/config.yml
```diff
@@ -49,6 +49,35 @@ commands:
...
+      - run: rm -rf *
+      - attach_workspace:
+          at: .
+      - run: touch .nojekyll
+      - run: git config --global user.email circleci@circleci
+      - run: git config --global user.name CircleCI
+      - run: git add *
+      - run: git commit -m "Add generated opera docs for GitHub pages"
+      - run: git push origin gh-pages
+
```
tadeboro, line 80: Doing the doc build in tow stages feels rather awkward. Maybe we can do it in a single step by utilizing git worktrees? This is how one would do that manually (translating this to the CircleCI YAML is left as an exercise for the reader ;)):

```
make docs
git worktree add gh-pages gh-pages
cd gh-pages
git rm -rf . )
cp -r ../docs/_build/* .
git add .
git comm -m "..."
git push origin gh-pages
```

.circleci/config.yml
```diff
@@ -49,6 +49,26 @@ commands:
...
       - run: git push origin stable
 
+  build_and_deploy_docs_to_github_pages:
+    description: Build Sphinx documentation and push GitHub stable branch which points to the latest relased tag/commit
+    steps:
+      - add_ssh_keys:
+          fingerprints:
+            - "b6:94:61:c3:e1:8f:25:e5:43:6d:51:9e:43:c4:16:11"
+      - common_setup
+      - checkout
```
tadeboro, line 59: Why do we need this? If I read things correctly, checkout is part of the common_setup step.

.circleci/config.yml
```diff
@@ -49,6 +49,26 @@ commands:
...
+      - run: make docs
+      - run: git worktree add gh-pages gh-pages
+      - run: cd gh-pages
```
tadeboro, line 62: This is a no-op since CircleCI executor is stateless. Any commands after this line are executed at the root of the checkout, making a mess.

anzoman: Aha, yes, I see that every new `run` starts a new process. But one possible solution we can use is the multi-line syntax for the run command which will prevent these commands from being forgotten.

.circleci/config.yml
anzoman: We don't need it so this has been removed.

.circleci/config.yml
anzoman: Thanks for the comment and suggestion, as a reader I tried to stick to your commands and translated them to CircleCI YAML entries.

anzoman:
@tadeboro I've refactored CI/CD and tested the pipeline. Everything seems to work fine but I think that we'll also need to exclude branch `gh-pages` from the CircleCI in order to get rid of build errors there.

sstanovnik:
I'd like to move this forward a bit. As the content is the same as we have in the xopera-docs repository, I have nothing to add there. There are just two questions:

* This only runs on the master branch, right? Just making sure I'm not missing anything.
* The process has been tested before limiting to the master branch, right? There are commits in `gh-pages` right now that seem to be exactly that.

anzoman:
@sstanovnik glad to here that!

- yes docs are build and published to GitHub Pages only on master branch
- yes, exactly, the test process that was limited to documentation branch succeeded and has created [github-pages environment](https://github.com/xlab-si/xopera-opera/deployments/activity_log?environment=github-pages) where the docs were deployed and published to https://xlab-si.github.io/xopera-opera/

sstanovnik:
Great. Let's just _go ahead_ and _pull the trigger_ on this one. Any changes can be addreses in a PR of a lower scope later. Dismissing the existing review and adding a new one for the sake of speed.

sstanovnik (approved):
see above

Issue #59, opened 2020-06-11
Add support for get_artifact TOSCA function


anzoman:
## Description
xOpera orchestration tool supports several [TOSCA functions](https://docs.oasis-open.org/tosca/TOSCA-Simple-Profile-YAML/v1.3/cos01/TOSCA-Simple-Profile-YAML-v1.3-cos01.html#_Toc26969452) at the moment such as `get_attribute`, `get_property`, `get_input` which are really useful. But we are still lacking the parser support for some of the TOSCA functions. Amongst them probably the most desired one would be the `get_artifact` TOSCA function which is used to retrieve an artifact location between TOSCA entities defined in the same TOSCA service template. So as a user I would definitely want to use my prepared orchestration artifacts (for instance zipped FaaS function files) directly in a service template by linking them with `get_artifact: [ <modelable_entity_name>, <artifact_name>, <location>, <remove> ]`. 

## Steps
To implement this we would need to update our parser and provide the support for this function in a similar way as it is provided for the currently supported TOSCA functions.

## Current behaviour
Currently xOpera does not recognize `get_artifact` function.

## Expected results
To be able to parse and use `get artifact` TOSCA function in templates like in the example below:

```yaml
node_templates:
  wordpress:
    type: tosca.nodes.WebApplication.WordPress
    ...
    interfaces:
      Standard:
        configure:
          create:
            implementation: wordpress_install.sh
            inputs:
              wp_zip: { get_artifact: [ SELF, zip ] }
    artifacts:
      zip: /data/wordpress.zip
```

anzoman:
The initial support for `get_artifact` TOSCA function was added with #67. It doesn't yet fully support all the possible options but it does work nicely and we will probably do some refactoring if needed. As adding this functionality was successful, the issue can be closed.

PR #67, opened 2020-06-23
Supplement get_artifact TOSCA function

anzoman:
Now the time is right to replenish our opera orchestrator with the
support for get_artifact TOSCA function which is generally used to
extract artifact location between modelable TOSCA entities that are
defined within the same service template. The desire for this useful
enhancement has also been the main topic of the issue #59.

The implemented TOSCA artifact function can be useful to link the
prepared orchestration artifacts (e.g. zipped FaaS function files)
directly to a TOSCA service template. Most likely the users will use
get_artifact function when setting the default value for interface
operation inputs and then later on retrieve them during the
orchestration in the Ansible playbook actuators on demand.

With these additions we supported all the parameters defined in TOSCA
specification that can be used in get_artifact (modelable_entity_name,
artifact_name, location and remove params). The defined artifacts need
to be at hand when referenced by TOSCA function and thus we copy them
to a temporary directory during the execution which is similar to
operation dependencies.
________________________________________________________________________________

This PR is still in its initial state and will probably need a lot of polishing before we can accept it. I did not implement all the `get_artifact` TOSCA function parameters like it's said in the commit description but I assumed that we will likely want to support them (or maybe not). The ones that have been left out are `location` and `remove` parameters that will need some extra attention because they interfere with the orchestration process itself. The definition of the `get_artifact` function from the [TOSCA YAML provile v1.3 spec](https://docs.oasis-open.org/tosca/TOSCA-Simple-Profile-YAML/v1.3/cos01/TOSCA-Simple-Profile-YAML-v1.3-cos01.html#_Toc26969460) is shown on the image below.

image

This PR also provides an example of `get_artifact` TOSCA function usage that can be found in `examples/artifacts` folder. It came in handy for the testing of the implementation using the commands below:

```console
(.venv) anzoman@ubuntu:~/Desktop/xopera-opera$ opera deploy examples/artifacts/service.yaml 
  Deploying artifacts_file_0
    Executing create on artifacts_file_0
(.venv) anzoman@ubuntu:~/Desktop/xopera-opera$ opera outputs
output_attr:
  description: Example of attribute output
  value: This is an example file content.
```

### DIFF
--- a/examples/artifacts/files/file.txt
+++ b/examples/artifacts/files/file.txt
@@ -1,0 +1,1 @@
+This is an example file content.
--- a/examples/artifacts/playbooks/create.yaml
+++ b/examples/artifacts/playbooks/create.yaml
@@ -1,0 +1,14 @@
+---
+- hosts: all
+  gather_facts: false
+
+  tasks:
+    - name: See what's in that linked file
+      command: "cat {{ file }}"
+      register: file_contents
+
+    - name: Set attribute
+      set_stats:
+        data:
+          my_attribute: "{{ file_contents.stdout }}"
+...
--- a/examples/artifacts/service.yaml
+++ b/examples/artifacts/service.yaml
@@ -1,0 +1,32 @@
+---
+tosca_definitions_version: tosca_simple_yaml_1_3
+
+node_types:
+  opera.nodes.file:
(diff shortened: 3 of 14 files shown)


cankarm (approved):
I approve.

Issue #68, opened 2020-07-03
Implement debug mode with vebose CLI option


anzoman:
## Description
Sometimes it's good to tell a little bit more. The same could apply for opera orchestrator where it would be useful to return more information during the deployment/undeployment process which would make the lives of users easier. This could be achieved by implementing an opera debug mode that could be used with every opera CLI command with and optional argument (for example `--verbose` or `--debug`). Opera would then provide and print out the additional info. 

The things that could be printed out are:

- actual values of input parameters
- return values of the executed Ansible playbooks
- the current values of the associated attributes after each operation 

and so on...

Example for this option would be:

```console
user@ubuntu opera deploy -i inputs.yaml service.yaml --verbose
Collected inputs:
    host_ip: localhost
    first: 1
    second: 2
    third: 3

Deploying my-workstation_0
Deploying hello-1
Deploying hello-2
  Executing create on hello-2

############################################################################################################################################################################
PLAY [localhost]

TASK [Install prerequsiste packages]
fatal: [localhost]: FAILED! => {"changed": false, "cmd": ["/home/.venv/bin/pip3", "install", "-U", "python>=2.6", "requests>=2.18.4", "google-auth>=1.3.0"], 
"msg": "stdout: Collecting python>=2.6\n\n:stderr:   Could not find a version that satisfies the requirement python>=2.6 (from versions: )\nNo matching distribution 
found for python>=2.6\nYou are using pip version 18.1, however version 20.2b1 is available.\nYou should consider upgrading via the 'pip install --upgrade pip' command.\n"}

PLAY RECAP
localhost                  : ok=0    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0
############################################################################################################################################################################

Deploying hello-3
Executing create on hello-3
...
```

## Steps
In order to implement the aforementioned behaviour we would have to add a new argument to all the opera's CLI commands and if user sets it we would have to call the debug version of the deployment/undeployment. 

## Current behaviour
As of now xOpera doesn't have verbose CLI argument.

## Expected results
To be able to use `--verbose`/`--debug` option with all opera CLI commands.

anzoman:
Opera now has `--verbose` or `-v` switch that turns on verbose mode, which is able to display orchestration inputs and the output of executed Ansible playbook tasks. Therefore, I think that we can close the issue.

PR #70, opened 2020-07-06
Include verbose mode for opera CLI commands

anzoman:
These changes apply to the #68 issue that expressed the need to give
users more information during the orchestration process. And because
less is not always more we decided to implement a debug mode for opera
which returns more information during the deployment/undeployment
process. This could was done by adding a new optional CLI argument
called --verbose or -v which can be used with opera deploy or opera
undeploy CLI commands. We also prepared verbose CLI option for
all the other remaining opera commands.

If the verbose mode is turned on user will be able to see the outputs
of the Ansible playbook executors and also orchestration inputs. This
will hopefully lead to easier debugging of the orchestration process
and especially the actuators of the orchestration.

Now the deploy/un-deploy opera commands would look like these:

```console
# deploy TOSCA template with higher verbosity level
opera deploy -i inputs.yaml service.yaml -v

# undeploy and turn on debug mode
opera undeploy --verbose
```

### DIFF
--- a/src/opera/commands/deploy.py
+++ b/src/opera/commands/deploy.py
@@ -29,6 +29,10 @@ def add_parser(subparsers):
             threads (positive number, default 1)",
         type=int, default=1
     )
+    parser.add_argument(
+        "--verbose", "-v", action='store_true',
+        help="Turns on verbose mode",
+    )
     parser.add_argument("template",
                         type=argparse.FileType("r"), nargs='?',
                         help="TOSCA YAML service template file",
@@ -67,7 +71,7 @@ def _parser_callback(args):
         return 1
 
     try:
-        deploy(service_template, inputs, storage, args.workers)
+        deploy(service_template, inputs, storage, args.verbose, args.workers)
     except ParseError as e:
         print("{}: {}".format(e.loc, e))
         return 1
@@ -78,7 +82,7 @@ def _parser_callback(args):
     return 0
 
 
-def deploy(service_template: str, inputs: typing.Optional[dict], storage: Storage, num_workers: int):
+def deploy(service_template: str, inputs: typing.Optional[dict], storage: Storage, verbose_mode: bool, num_workers: int):
     """
     :raises ParseError:
     :raises DataError:
@@ -94,4 +98,4 @@ def deploy(service_template: str, inputs: typing.Optional[dict], storage: Storag
     ast = tosca.load(Path.cwd(), PurePath(service_template))
     template = ast.get_template(inputs)
     topology = template.instantiate(storage)
-    topology.deploy(num_workers)
+    topology.deploy(verbose_mode, num_workers)
--- a/src/opera/commands/undeploy.py
+++ b/src/opera/commands/undeploy.py
@@ -22,6 +22,10 @@ def add_parser(subparsers):
             threads (positive number, default 1)",
         type=int, default=1
     )
+    parser.add_argument(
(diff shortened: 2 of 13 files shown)


src/opera/commands/deploy.py
```diff
@@ -45,6 +49,8 @@ def deploy(args):
         print("{0} is not a positive number!".format(args.workers))
         return 1
 
+    verbose_mode = True if args.verbose else False
```
tadeboro, line 52: `args.verbose` is already a boolean, so no need for conversion here.

src/opera/commands/undeploy.py
```diff
@@ -32,6 +36,7 @@ def undeploy(args):
     storage = Storage(Path(args.instance_path)) if args.instance_path else Storage(Path(".opera"))
     root = storage.read("root_file")
     inputs = storage.read_json("inputs")
+    verbose_mode = True if args.verbose else False
```
tadeboro, line 39: Same issue as in the deploy parser.

src/opera/executor/ansible.py
```diff
@@ -27,7 +27,7 @@ def _get_inventory(host):
     return yaml.safe_dump(dict(all=dict(hosts=dict(opera=inventory))))
 
 
-def run(host, primary, dependencies, vars):
+def run(host, primary, dependencies, vars, verbose=False):
```
tadeboro, line 30: Is there a reason behind the default value here? If we are carrying this parameter over several layers of the app, it would be wiser to make it required to ensure we did not forget to parameterize any of the places that need verbosity adjusted.

src/opera/commands/deploy.py
anzoman: Ok, I've removed that.

src/opera/commands/undeploy.py
anzoman: This was also removed.

src/opera/executor/ansible.py
anzoman: Yeah, I agree. So I have removed default False values for verbose function parameter so this should make it to be required.

sstanovnik:
This is something I found myself wanting as well. I'd accept this PR as it is now (but with change 1 below applied), but then make multiple changes to the whole thing. In separate PRs of course, things would get bloated quickly.

1. Apply the --verbose flag to _all_ commands, even if it does nothing. Consistency is very important.
2. I'd really really prefer opera-global flags, such as --verbose and --instance-path to be in the global parser. Ideally `opera deploy -v` would work with that, but I'd accept `opera -v deploy` to be the only acceptable option. For now. It bothers me a bit when I have to be careful of where to position my flag parameters.
3. Print statements are _awful_ and I found myself hating them a lot in xopera-api. I had to resort to an ugly hack to capture stdout. How about a logger? As in `import logging`?
4. The verbosity parameter should really only be applied to the logger. No need to pass the parameter downstream to anything, pass the logger, and you don't need to be careful of missing any. If you call `log.debug()` it's one level, `log.info()` is another.

Well, this is my wishlist. As I said, this is a big step in the right direction, but there are many ways this could be made even better!

anzoman:
@sstanovnik thanks for your comments and wishes! To remain consistent I've added verbose CLI option also to the other opera commands (validate, outputs, init). And as you said this is just the beginning.

sstanovnik (approved):
All of my and @tadeboro's comments have been addressed, and this looks nice.

PR #71, opened 2020-07-07
Add new opera deploy cli command options

### LINKED ISSUE #46: Add `deploy -f <>` option
## Description
Opera saves the deployment process output and therefore it can resume deploy procedure which was not successful or it ended prematurely. In some cases the DevOps engineer would prefer to re-run everything from the start and not skipping successful Ansible script executions. The limits or potential problems of this procedure are described in NOTE section below.

## Prerequisites
- `opera` installed
- project, e.g. `service.yaml` 

## Steps

- User starts deploy, which is interrupted

```
$ opera deploy service.yaml 
  Deploying first node
     Executing create on the_first_ansible_role
  Deploying second node
Interrupt !!!

```
- continue without force:

```
$ opera deploy service.yaml 
  Deploying second node
    Executing create on the_second_ansible_role
```

- **continue with force**
 ```
$ opera deploy -f service.yaml 
  Deploying first node
     Executing create on the_first_ansible_role
  Deploying second node
    Executing create on the_second_ansible_role
```

## Expected results
- continue with force should re-run all steps and Ansible playbooks.

## NOTE
This action should be available only if application blueprint did not changed its structure, otherwise some deployed nodes from previous deployment would become un-managed by xOpera, and would be impossible to undeploy/clean them if they are not described in a _new_ blueprint. For a start I would suggest to allow this force command only if blueprint did not change at all.

If user creates significant changes in blueprint, user should clean previous deployment, e. g.:
  ```
$ opera deploy -f service.yaml 
   Error: The service.yaml blueprint structure has changed. You should run "opera undeploy" first to clean and re-deploy service, or use unchanged blueprint. 
```

anzoman:
The discussion in #46 showed that it would be necessary to add some
optional arguments to opera deploy command. In this commit we did just
that in order to address the new features which are:

```console
# deploy (start from the beginning, fail when already deployed)
opera deploy -i inputs.yaml service.yaml

# resume (start from the point of failure)
opera deploy --resume -i inputs.yaml service.yaml

# force (start over)
opera deploy --force -i inputs.yaml service.yaml
```

a) Wise men once said that TOSCA orchestrator should treat each
deployment separately and distinguish it from all other deployments
that have been/would be created before/after. And because opera treats
deployments based on file system storage location (i.e. hidden .opera
folder) we went on and implemented a different deployment behaviour.
So if user tries to deploy again on already deployed instance model an
error is raised telling that deployment has already been done.

b) On the other hand users would often like to continue where they left
or where their deployment process has been interrupted (for example
because of some error in Ansible playbook) and this would then skip all
the instances that have been deployed. To realize that we added
--resume or -r option where user can continue his orchestration.

c) And lastly DevOps engineers would sometimes prefer to re-run the
whole deployment from the starting point without skipping the tasks
that were successfully deployed. For that cases --force or -f optional
CLI argument was introduced which erases the previous instance model
and create a new, fresh one. And to have no worries we ensured that
--resume and --force options are mutually exclusive.
________________________________________________________________

The remaining commit messages in this PR are based on #55 which is ready to be merged. This also means that the changes work when using `opera init` command  before deploying so that user can do `opera deploy {--force|--resume}`.
________________________________________________________________

Resolves #46

### DIFF
--- a/src/opera/commands/deploy.py
+++ b/src/opera/commands/deploy.py
@@ -5,6 +5,7 @@ from pathlib import Path, PurePath
 
 import yaml
 
+from opera.utils import prompt_yes_no_question
 from opera.error import DataError, ParseError
 from opera.parser import tosca
 from opera.storage import Storage
@@ -21,13 +22,28 @@ def add_parser(subparsers):
     )
     parser.add_argument(
         "--inputs", "-i", type=argparse.FileType("r"),
-        help="Optional: YAML file with inputs to override the inputs supplied in init",
+        help="Optional: YAML file with inputs to "
+             "override the inputs supplied in init",
     )
     parser.add_argument(
         "--workers", "-w",
-        help="Maximum number of concurrent deployment \
-            threads (positive number, default 1)",
+        help="Maximum number of concurrent deployment "
+             "threads (positive number, default 1)",
         type=int, default=1
+    )
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument(
+        "--resume", "-r", action='store_true',
+        help="Resume the deployment from where it was interrupted",
+    )
+    group.add_argument(
+        "--clean-state", "-c", action='store_true',
+        help="Clean the previous deployment state "
+             "and start over the deployment",
+    )
+    parser.add_argument(
+        "--force", "-f", action='store_true',
+        help="Force the action and skip any possible prompts",
     )
     parser.add_argument(
         "--verbose", "-v", action='store_true',
@@ -42,7 +58,8 @@ def add_parser(subparsers):
 
 def _parser_callback(args):
     if args.instance_path and not path.isdir(args.instance_path):
-        raise argparse.ArgumentTypeError("Directory {0} is not a valid path!".format(args.instance_path))
+        raise argparse.ArgumentTypeError("Directory {0} is not a valid path!"
+                                         .format(args.instance_path))
 
     storage = Storage.create(args.instance_path)
 
@@ -50,15 +67,40 @@ def _parser_callback(args):
         print("{0} is not a positive number!".format(args.workers))
         return 1
 
-    # TODO(@tadeboro): This should be part of the init command that we do not
-    # have yet.
+    delete_existing_state = False
+    if storage.exists("instances"):
+        if args.resume:
+            if not args.force:
+                print("The resume deploy option might have unexpected "
+                      "consequences on the already deployed blueprint.")
+                question = prompt_yes_no_question()
+                if not question:
+                    return 0
+        elif args.clean_state:
+            if args.force:
+                delete_existing_state = True
+            else:
+                print("The clean state deploy option might have unexpected "
+                      "consequences on the already deployed blueprint.")
+                question = prompt_yes_no_question()
+                if question:
+                    delete_existing_state = True
+                else:
+                    return 0
+        else:
+            print("The instance model already exists. Use --resume to "
+                  "continue or --clean-state to delete current deployment "
+                  "state and start over the deployment.")
+            return 0
+
     if args.template:
         service_template = args.template.name
     else:
         if storage.exists("root_file"):
             service_template = storage.read("root_file")
         else:
-            print("CSAR or template root file does not exist. Maybe you have forgotten to initialize it.")
+            print("CSAR or template root file does not exist. "
+                  "Maybe you have forgotten to initialize it.")
             return 1
 
     try:
@@ -71,7 +113,8 @@ def _parser_callback(args):
         return 1
 
     try:
-        deploy(service_template, inputs, storage, args.verbose, args.workers)
+        deploy(service_template, inputs, storage, args.verbose, args.workers,
+               delete_existing_state)
     except ParseError as e:
         print("{}: {}".format(e.loc, e))
         return 1
@@ -82,11 +125,16 @@ def _parser_callback(args):
     return 0
 
 
-def deploy(service_template: str, inputs: typing.Optional[dict], storage: Storage, verbose_mode: bool, num_workers: int):
+def deploy(service_template: str, inputs: typing.Optional[dict],
+           storage: Storage, verbose_mode: bool, num_workers: int,
+           delete_existing_state: bool):
     """
     :raises ParseError:
     :raises DataError:
     """
+    if delete_existing_state:
+        storage.remove("instances")
+
     if inputs is None:
         if storage.exists("inputs"):
             inputs = yaml.safe_load(storage.read("inputs"))
--- a/src/opera/commands/init.py
+++ b/src/opera/commands/init.py
@@ -15,7 +15,8 @@ from opera.storage import Storage
 def add_parser(subparsers):
     parser = subparsers.add_parser(
         "init",
-        help="Initialize the deployment environment for the service template or CSAR"
+        help="Initialize the deployment environment "
+             "for the service template or CSAR"
     )
     parser.add_argument(
         "--instance-path", "-p",
@@ -37,7 +38,8 @@ def add_parser(subparsers):
 
 def _parser_callback(args):
     if args.instance_path and not path.isdir(args.instance_path):
-        raise argparse.ArgumentTypeError("Directory {0} is not a valid path!".format(args.instance_path))
+        raise argparse.ArgumentTypeError("Directory {0} is not a valid path!"
+                                         .format(args.instance_path))
 
     storage = Storage.create(args.instance_path)
 
@@ -67,7 +69,9 @@ def _parser_callback(args):
     return 0
 
 
-def initialize_compressed_csar(csar_name: str, inputs: typing.Optional[dict], storage):
+def initialize_compressed_csar(csar_name: str,
+                               inputs: typing.Optional[dict],
+                               storage: Storage):
     if inputs is None:
         inputs = {}
     storage.write_json(inputs, "inputs")
@@ -91,7 +95,9 @@ def initialize_compressed_csar(csar_name: str, inputs: typing.Optional[dict], st
     template.instantiate(storage)
 
 
-def initialize_service_template(service_template: str, inputs: typing.Optional[dict], storage: Storage):
+def initialize_service_template(service_template: str,
+                                inputs: typing.Optional[dict],
+                                storage: Storage):
     if inputs is None:
         inputs = {}
     storage.write_json(inputs, "inputs")
--- a/src/opera/commands/outputs.py
+++ b/src/opera/commands/outputs.py
@@ -41,7 +41,8 @@ def format_outputs(outputs, format):
 
 def _parser_callback(args):
     if args.instance_path and not path.isdir(args.instance_path):
-        raise argparse.ArgumentTypeError("Directory {0} is not a valid path!".format(args.instance_path))
+        raise argparse.ArgumentTypeError("Directory {0} is not a valid path!"
+                                         .format(args.instance_path))
 
     storage = Storage.create(args.instance_path)
     try:
@@ -66,6 +67,7 @@ def outputs(storage: Storage) -> dict:
 
     ast = tosca.load(Path.cwd(), PurePath(service_template))
     template = ast.get_template(inputs)
-    # We need to instantiate the template in order to get access to the instance state.
+    # We need to instantiate the template in order
+    # to get access to the instance state.
     template.instantiate(storage)
     return template.get_outputs()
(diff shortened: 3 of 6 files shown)


anzoman:
cc @cankarm and @dradx just to let you know that things are moving forward here. If you have any new suggestions considering this PR (and our discussion in #46) you can also express them here.

src/opera/commands/deploy.py
```diff
@@ -28,6 +28,14 @@ def add_parser(subparsers):
             threads (positive number, default 1)",
         type=int, default=1
     )
+    parser.add_argument(
+        "--resume", "-r", action='store_true',
+        help="Resume the deployment from where it was interrupted.",
+    )
+    parser.add_argument(
+        "--force", "-f", action='store_true',
+        help="Delete the previous instance model and start over the deployment.",
```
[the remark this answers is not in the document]
anzoman: Here with `--force` option we are deleting the instance model that gets created with `opera deploy` and is present in `.opera/instances` folder. With these files gone the deployment can be started again without un-deploying anything. So if we want to go this way `opera redeploy` could be introduced as the new command that would undeploy the instances first and then it would deploy them again.

src/opera/commands/deploy.py
```diff
@@ -45,8 +57,23 @@ def deploy(args):
...
+                return 0
+        elif args.force:
+            print("The force deploy option might have unexpected consequences on the already deployed blueprint.")
+            question = prompt_yes_no_question()
+            if question:
+                storage.remove("instances")
+            else:
+                return 0
+        else:
+            print("The instance model already exists. To redeploy use --force option or --resume to continue.")
```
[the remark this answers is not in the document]
anzoman: Yes, good call I will expand the message a little bit. The third option you mentioned can be skipped for now since `redeploy` may be the new command that will need to be implemented.

src/opera/commands/deploy.py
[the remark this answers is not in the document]
cankarm: No need for a new "redeploy" action. Even though the message is quite clear, I would vote for changing it to

"Clear orchestrator metadata explaining the state of the deployment (instance model) and start over the deployment process. Already deployed instances will not be undeployed."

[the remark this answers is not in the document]
matejart: I actually like @anzoman 's original message better. At best, if you feel "instance model" is not something the users would understand (even though it is a term in the standard), you can go for "Delete the previous deployment state ..." or "Abandon deployment state and start over ...".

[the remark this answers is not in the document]
cankarm: I would prefer to emphasise "state". Just because the initial sentence is too similar to the "Delete the previous instance and start over...".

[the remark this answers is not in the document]
anzoman: Ok, I've refactored the message to `Delete the previous deployment state and start over the deployment.`

src/opera/commands/deploy.py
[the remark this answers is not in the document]
anzoman: The new message is: "The instance model already exists. Use --resume to continue or --force to ignore current deployment state and start over the deployment."

cankarm:
I don't have further comments, LGTM. 

@dradX?

anzoman:
Thanks @cankarm, it is also good to mark that when this PR is merged it will also fully resolve #46.

src/opera/commands/deploy.py
```diff
@@ -37,24 +47,50 @@ def add_parser(subparsers):
...
+            if question:
+                storage.remove("instances")
+            else:
+                return 0
+        else:
+            print("The instance model already exists. Use --resume to "
+                  "continue or --force to ignore current deployment state "
+                  "and start over the deployment.")
+            return 0
+
```
sstanovnik, line 85: This whole block needs to be restructured to reside in `deploy()`, because it's application logic, not argument validation and parsing. A good guideline is that if the CLI is the only one that could possibly do this, it goes into the callback, otherwise it needs to be in the library method.

Here, removing the instances file needs to be in the library function, but the prompts mustn't be there. You need to think about how you want to design the CLI and library - here is my idea:

* Make the CLI user choose between resuming and forcing.
  * Resuming ignores any errors, forcing deletes the instance file (==sets a flag, see next line)
* Make the library user choose whether to delete a previous deployment by adding a `delete_existing` boolean parameter. This makes the library user explicitly choose whether to silently ignore an existing deployment or delete it.

The only thing you need to do (to keep me happy :)) is to keep the library interface usable, and here that includes not having special behaviour for the CLI.

src/opera/commands/deploy.py
```diff
@@ -21,14 +21,24 @@ def add_parser(subparsers):
...
         type=int, default=1
     )
+    parser.add_argument(
+        "--resume", "-r", action='store_true',
+        help="Resume the deployment from where it was interrupted.",
+    )
+    parser.add_argument(
+        "--force", "-f", action='store_true',
+        help="Delete the previous deployment state "
+             "and start over the deployment.",
```
sstanovnik, line 40: Force is a bit awkward for me. As soon as I saw the prompts below, I thought --force just skipped them. But no, it deletes the previous deployment. How about `-d/--delete-existing`?

src/opera/commands/deploy.py
```diff
@@ -95,3 +132,18 @@ def deploy(service_template: str, inputs: typing.Optional[dict], storage: Storag
...
+def prompt_yes_no_question():
+    check = str(input("Do you want to continue? (Y/n): ")).lower().strip()
+    try:
+        if check[:1] == "" or check[:1] == 'y' or check[:1] == "yes":
+            return True
+        elif check[:1] == 'n' or check[:1] == "no":
+            return False
+        else:
+            print('Invalid input. Please try again.')
+            return prompt_yes_no_question()
```
sstanovnik, line 146: Maybe use `str.startswith()` here? Much clearer as to what's really going on at first glance. Also have a list of yes and a list of no options? Something like this:

```python3
def prompt_yes_no_question(yes_responses=("y", "yes"),
                           no_responses=("n", "no"),
                           case_sensitive=False,
                           default_response="yes"):
   ...
```

src/opera/commands/deploy.py
```diff
@@ -95,3 +132,18 @@ def deploy(service_template: str, inputs: typing.Optional[dict], storage: Storag
     template = ast.get_template(inputs)
     topology = template.instantiate(storage)
     topology.deploy(num_workers)
+
+
+def prompt_yes_no_question():
```
sstanovnik, line 137: This really seems like something that should be in a separate file, maybe `opera/commands/__init__.py`? So you'd `from opera.commands import prompt_yes_no_question`

But then you'd run into circular imports, so maybe a utils file?

src/opera/commands/deploy.py
```diff
@@ -37,24 +47,50 @@ def add_parser(subparsers):
 
 
 def _parser_callback(args):
+    if args.resume and args.force:
+        print("The -r/--resume and -f/--force "
+              "options cannot be used together.")
+        return 0
```
sstanovnik, line 53: This is a feature of the parser and would be better implemented like that: https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group

src/opera/commands/deploy.py
anzoman: Yeah, maybe force can be confusing from the perspective of those who look into the code, but for opera CLI users probably not since they would force the deployment. What do you think @matejart, @cankarm, @dradx?

sstanovnik: But I'd expect "forcing" to just do what it did previously, before these checks were introduced. Maybe not though, I'm really not a user of xopera :)

anzoman: Before this PR we had no forcing and if you tried to deploy again on already deployed instance nothing really happened. But `-d/--delete-existing` could be if everyone thinks it's ok.

cankarm: @sstanovnik The root of this issue lies in statefulness of xOpera. The orchestrator saves the states and if you deployed something, then you do a manual change on deployed infra or maybe a small change of some module config, another xopera deploy would ignore that. Deploy would check the state written in .opera and skip all successful steps.

And then there is an argue what is _normal expected_ behaviour for `deploy`. Should always start to redeploy something regardless of the state in .opera or skip successful steps and just resume... this long discussion is in issue #46 . We decided to use switches that exactly explain what happens.

So `-d/--delete-existing` seems promising, but what exactly would be, delete existing. This force actually deletes/ignores only opera state (so .opera) and not "delete deployed instances". This only means that ansible scripts will be re-run on all instances.

src/opera/commands/deploy.py
cankarm: Indeed, the functionality should be also available for the xOperaAPI and not only for CLI. Nice catch @sstanovnik.

src/opera/commands/deploy.py
anzoman: Thanks, now this is included.

src/opera/commands/deploy.py
dradX: I would agree with this implementation by introducing `--force/-f `  and `--resume/-r` - these are nice features, but as I pointed out in a previous comment I really don't see any point of prompting users after he/she expresses a clear request by using these switches. 

Otherwise I agree with @sstanovnik that this features should be made available in API (library) calls. The `--resume/-r` seems especially interesting in HPC long running jobs scenarios - when something goes wrong and the user wishes not to repeat the successfully executed jobs.

src/opera/commands/deploy.py
anzoman: I see. For now I've created utils.py file where prompt function resides.

src/opera/commands/deploy.py
anzoman: Ok, I've refactored this and used almost similar signature as you suggested. I don't think we need `str.startswith()` because we are now searching for the input in list of yes/no options.

src/opera/commands/deploy.py
anzoman: Removing instances has been moved to the library method and prompts and CLI arguments were kept where they had been before. I've also used the `delete_existing` boolean parameter.

cankarm: and what delete existing will do?

a) full undeploy and delete .opera ?
b) only delete .opera temp folder?

anzoman: `delete_existing` would delete `instances` subfolder that is contained in `.opera` folder which means that it would erase the created instance model.

cankarm: my personal problem is that `delete_existing` is not self-explanatory. Probably we already had an idea of  `delete_state` or  remove the current state with `clean` or `state clean`, etc.

anzoman: Aha I see. No worries - the parameter can simply be renamed to `delete_instances` or `delete_instance_model` or `delete_state` to make it clearer.  `opera deploy --force` deletes the created instance model and starts the deployment again. However, if we would also like to have a new argument (i.e. `--clean-state` like you mentioned), which would only delete instance model (i.e. state) we can do that in a this/separate PR but the usage of this clean option then needs to be explained further ad more in detail.

anzoman:
The latest CLI options we have are:

```console
# deploy (start from the beginning, warning when already deployed)
opera deploy -i inputs.yaml service.yaml

# resume (start from the point of failure)
opera deploy --resume/-r -i inputs.yaml service.yaml

# clean-state (clean instances folder and start over the deployment)
opera deploy --clean-state/-c -i inputs.yaml service.yaml

# force option skips prompts and warnings
opera deploy -r/-c --force/-f -i inputs.yaml service.yaml
```

@sstanovnik, can you take a look? :)

sstanovnik:
The new commands are much clearer as to what they do, yes. A couple of questions:

* there are merge conflicts, please rebase
* if a previous deployment exists, and you `opera deploy service.yml`, you issue an error, not a warning, right?
* if I do _not_ have an existing deployment and do `--resume`, do I still get prompted?

I could read the code thoroughly to answer the last two, but I think it's better that this gets discussed as an explicit decision :)

anzoman:
@sstanovnik to answer you questions:

- merge conflicts are now resolved
- yes, result is an error saying: `The instance model already exists. Use --resume to continue or --clean-state to delete current deployment state and start over the deployment.`
- no you don't because we first check if an instance model already exists, and only then `--resume/-r` can be applied and used

sstanovnik (approved):

sstanovnik:
Great, this now looks very much usable. If there are any changes for uses we haven't thought of, let's just decide on that later. :thumbs-up:

PR #75, opened 2020-07-15
Add parsing of TOSCA Policy type triggers

anzoman:
The desire for the support of [TOSCA Trigger definitions](https://docs.oasis-open.org/tosca/TOSCA-Simple-Profile-YAML/v1.3/os/TOSCA-Simple-Profile-YAML-v1.3-os.html#DEFN_ENTITY_TRIGGER_DEFN) has grown
stronger with opening the #32 issue. At this point we are therefore
introducing the support for parsing TOSCA Policy type triggers which
can be used to define the event, condition and action that is used to
invoke(trigger) a policy it is associated with. These changes also
include unit tests and the runnable TOSCA template example in
`examples/policy_triggers` that contains policies with TOSCA triggers. 
In the implementation part we first targeted to support the standard
trigger definition keynames, but then continued adding the support for
the additional keynames that are used in extended trigger notation.

With these changes we also refactor parsing policy type targets and add
new TOSCA entities like activity, condition clause and event filter, time 
interval etc. along with all the tests and examples.

### DIFF
--- a/examples/policy_triggers/playbooks/auto_scale.yaml
+++ b/examples/policy_triggers/playbooks/auto_scale.yaml
@@ -1,0 +1,8 @@
+---
+- hosts: all
+  gather_facts: false
+
+  tasks:
+    - name: Autoscale tasks
+      debug:
+        msg: Insert tasks for auto-scaling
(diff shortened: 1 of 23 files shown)


cankarm (approved):
This is a huge PR including a lot of tests. I'll aprove this, so the functionality can be merged asap.

PR #80, opened 2020-07-24
Increase verbosity when encountering unset values

anzoman:
Up until now opera printed "Cannot use an unset value." in case there
was an input defined in the service template but was not set within
inputs.yaml file. There are also other cases for example when you
define a property or attribute within node type without setting its
default value. If you forget to define its value later within the
node_templates section you will get the same error. And since it would
be really useful to know which property/attribute/input causes an issue
in such cases when the data evaluation fails we decided to give more
information to the users by printing also the name of the property.

### DIFF
--- a/src/opera/instance/node.py
+++ b/src/opera/instance/node.py
@@ -130,7 +130,7 @@ class Node(Base):
         # TODO(@tadeboro): Add support for nested attribute values once we
         # have data type support.
         if attr in self.attributes:
-            return self.attributes[attr].eval(self)
+            return self.attributes[attr].eval(self, attr)
 
         # Check if there are capability and requirement with the same name.
         if attr in self.out_edges and attr in [c.name for c in
--- a/src/opera/instance/relationship.py
+++ b/src/opera/instance/relationship.py
@@ -30,7 +30,7 @@ class Relationship(Base):
         # have data type support.
         if attr not in self.attributes:
             raise DataError("Instance has no '{}' attribute".format(attr))
-        return self.attributes[attr].eval(self)
+        return self.attributes[attr].eval(self, attr)
 
     def get_property(self, params):
         host, prop, *rest = params
--- a/src/opera/template/node.py
+++ b/src/opera/template/node.py
@@ -58,7 +58,9 @@ class Node:
         ), None)
         if host:
             instance = next(iter(host.instances.values()))
-            return instance.attributes["public_address"].eval(self)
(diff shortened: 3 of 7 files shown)


anzoman:
@sstanovnik one simple example that could prove this PR worthy should be for instance this one:

```yaml

node_types:
  hello_type:
    interfaces:
      Standard:
        inputs:
          marker:
            default: { get_input: marker }
            type: string
        operations:
          create: playbooks/create.yaml
          delete: playbooks/delete.yaml

topology_template:
  inputs:
    marker:
      type: string
# if we do not provide the value for the input
#      default: default-marker

  node_templates:
    my-workstation:
      type: tosca.nodes.Compute
      attributes:
        private_address: localhost
        public_address: localhost

    hello:
      type: hello_type
      requirements:
        - host: my-workstation
```

As you can see this is the same TOSCA service template as the one that resides in /examples/hello/service.yaml but without the default `marker` input value. So if we run this with `opera deploy service.yaml` we get: `Error processing node hello_0: Cannot use an unset value.`. With the changes from this PR we would get: `Error processing node hello_0: Cannot use an unset value: marker` so we could now where this unset value originates from. Here I demonstrated this with inputs but the same applies for property and attribute values.

cankarm (dismissed):

anzoman:
@sstanovnik and @cankarm right now we are using static variable to store evaluated parameters of the TOSCA functions (get_artifact, get_property, get_input,...) which, if unset, are then printed out along with the error message. And I have not found any good idea on how to improve this right now.

cankarm:
@anzoman, would we need to improve it currently. I'm not sure that I get what do you suggest. Would you like some ideas on how to improve this?

anzoman:
@cankarm the functionality is okay and everything works nicely. My previous comment was targeting the improvements for the code which was discussed with @sstanovnik. Here we use a static variable to detect unset values, which is not the best solution since the variable can be easiliy modified when witing new code and that could break the functionality.

cankarm:
Ok, thanks for the clarification!

anzoman:
I created an update for this PR. So previously we used a static variable to store unset values and we applied the following changes:

image

Now the story is different as we use one additional function argument for that. What do you think @sstanovnik?

sstanovnik (approved):
This looks much better! While I'm not sure if this is optimal because I don't know the parser, there aren't any obvious code smells either, so it's perfect.

Issue #81, opened 2020-07-24
opera init: `create.yml` does not exist


duelle:
Version: `opera 0.5.8`

When running the `opera init http.zip` command with the provided zip file ([http.zip](https://github.com/xlab-si/xopera-opera/files/4973145/http.zip)), the following error is reported:

```
% opera init http.zip 
.opera/csars/csar/_definitions/radonnodestesting__DeploymentTestAgent.tosca:23:24: [Path] Path /tmp/http/nodetypes/radon.nodes.testing/DeploymentTestAgent/files/create/create.yml does not exist.
```

The file is present in the CSAR/ZIP file but obviously not in the given path as the extraction seems to happen inside the `.opera` folder and not directly to the folder the  `http.zip` file is located in (`/tmp/http/`).

The `create.yml` is available in the `/tmp/http/.opera/csars/csar/nodetypes/radon.nodes.testing/DeploymentTestAgent/files/create` folder after the `init` command though.

anzoman:
HI @duelle thanks for spotting this issue. I tried to run this myself and I get the same result. 

This seems to be firstly because of the working directory you were running the opera commands from. Like if the CSAR would be extracted to `/tmp/http/csar` and then if you run it from `tmp/http` you would get the same error because you are of course not running from the root of the csar. So when you move to `tmp/http/csar` everything works fine. But there is a catch with compressed csars which are (as you found out) extracted to `.opera/csars`. Currently only one CSAR can be used within the same working directory and it gets extracted to `.opera/csars/csar` (so it it also renamed to csar). But when you try to initialized it everything would work if you had your playbooks deeper in the structure than the TOSCA service template. However, the problem occurs when you want to link your playbooks from the root of the CSAR because opera assumes that where your current dir is, there must also be the root of the CSAR you're trying to initialize/deploy.  So if you are in `tmp/http` and you are running `opera init http.zip` opera searches for the linked playbooks in the current working dir but not in the actual root of the CSAR (`.opera/csars/csar`) which would be right.

So in order to overcome this issue we will have to refactor some code so that opera will use `.opera/csars/csar` as the working directory when initializing/deploying/undeploying compressed CSAR files. Do you agree @matejart?

anzoman:
This has been fixed with #82. Now when we try to deploy the [http.zip](https://github.com/xlab-si/xopera-opera/files/4973145/http.zip) CSAR, opera proceeds without any errors:

```console
[...]

PR #82, opened 2020-07-27
Deploy initialized compressed CSARs from their extracted root

anzoman:
The support for the compressed CSAR files was introduced along with
opera init command in #55. Currently one CSAR can be initialized within
the same working directory and it gets extracted to .opera/csars/csar.
The orchestration processed including CSAR initialization, deployment
and undeployment worked just fine when the linked artifacts (e.g
Ansible playbooks) for TOSCA operation implementations were deeper in
the structure than the runnable TOSCA service template. However, the
problem has arisen when user wanted to link his playbooks from the root
of the CSAR because opera assumed that the current working directory
that CSAR was deployed from, was at the root of the extracted CSAR ,
Till now opera was searching for the linked files in the current
working dir but not in the actual root of the CSAR (.opera/csars/csar)
which would be necessary. To overcome this issue we have refactored the
opera commands so that opera will use extracted CSAR root storage
folder .opera/csars/csar as the working directory when initializing,
deploying or undeploying compressed CSAR files. The changes we are
applying here are linked with the issue #81 where all this was spotted.

### DIFF
--- a/src/opera/commands/deploy.py
+++ b/src/opera/commands/deploy.py
@@ -1,6 +1,6 @@
 import argparse
 import typing
-from os import path
+from os import chdir, path
 from pathlib import Path, PurePath
 
 import yaml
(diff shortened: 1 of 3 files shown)


cankarm (approved):
I approve.

PR #85, opened 2020-07-31
Decouple the Python API from CLI interaction.

sstanovnik:
This will allow clients, e.g. xopera-api, to use this as a nice Python library. I've extracted the actual logic to separate functions and used a different one for CLI parser callbacks. Much cleaner this way :)

### DIFF
--- a/src/opera/commands/deploy.py
+++ b/src/opera/commands/deploy.py
@@ -1,4 +1,6 @@
 import argparse
+import typing
+from os import path
 from pathlib import Path, PurePath
 
 import yaml
@@ -6,7 +8,6 @@ import yaml
 from opera.error import DataError, ParseError
 from opera.parser import tosca
 from opera.storage import Storage
-from os import path
 
 
 def add_parser(subparsers):
@@ -32,14 +33,14 @@ def add_parser(subparsers):
                         type=argparse.FileType("r"), nargs='?',
                         help="TOSCA YAML service template file",
                         )
-    parser.set_defaults(func=deploy)
+    parser.set_defaults(func=_parser_callback)
 
 
-def deploy(args):
+def _parser_callback(args):
     if args.instance_path and not path.isdir(args.instance_path):
         raise argparse.ArgumentTypeError("Directory {0} is not a valid path!".format(args.instance_path))
 
-    storage = Storage(Path(args.instance_path)) if args.instance_path else Storage(Path(".opera"))
+    storage = Storage.create(args.instance_path)
 
     if args.workers < 1:
         print("{0} is not a positive number!".format(args.workers))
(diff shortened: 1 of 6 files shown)


src/opera/commands/deploy.py
```diff
@@ -1,12 +1,13 @@
 import argparse
+import typing
```
anzoman, line 2: I like typing. But do you think that this should break something?

src/opera/commands/deploy.py
```diff
@@ -85,3 +77,21 @@ def deploy(args):
         return 1
 
     return 0
+
+
+def deploy(service_template: str, inputs: typing.Optional[dict], storage: Storage, num_workers: int):
+    """
+    :raises ParseError:
```
anzoman, line 84: Is this necessary? Or what does this differ from try-except?

src/opera/storage.py
```diff
@@ -3,6 +3,12 @@
 
 
 class Storage:
+    DEFAULT_INSTANCE_PATH = ".opera"
+
+    @classmethod
+    def create(cls, instance_path: str = None) -> "Storage":
```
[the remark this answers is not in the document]
sstanovnik: This is a python quirk. If you want to add a type hint that references a class at the class level, you need to quote it, because the class name doesn't exist yet. Slightly strange that it wouldn't exist as we're obviously inside the class already, but that's the way the parser works apparently.

src/opera/commands/deploy.py
sstanovnik: `typing` and the type hints I used are available since python 3.6 - and I've checked that the PyPI page lists Python 3.6 and 3.7, while tests use 3.6, 3.7 and 3.8. I can move this typing information to comments if you want to preserve Python 3.5 compatibility, if there is any.

Python 3.5 also [EOLs on 2020-09-13](https://devguide.python.org/#status-of-python-branches). :)

src/opera/commands/deploy.py
sstanovnik: This doesn't replace try-except, it just documents that the function can throw these two errors. There aren't any checked exceptions in Python, so you never know what anything can throw, but this is a nice way of signalling to userss of opera as a library of what can go wrong. You can catch something better than `Exception`.

src/opera/storage.py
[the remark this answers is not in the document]
anzoman: Aha, interesting!

src/opera/commands/deploy.py
anzoman: Ok, then I think we are totally cool with that :)

src/opera/commands/deploy.py
anzoman: Great, that's ok then. And thanks for the explanation.

PR #90, opened 2020-08-12
Use occurrences keyname to control usage of requirements

### LINKED ISSUE #88: Raise an exception when targeting more than one instance via the same requirement
## Description
When defining node type's requirements as capabilities it is only possible to create the relationship between two nodes via the same requirement within the same node template.  Sometime users would want to connect more than one nodes (of different types) to the same node within the same requirement. However, this cannot survive through TOSCA template deployment since the relationships between nodes must be defined by using different requirements. For example in the TOSCA template below we try to link nodes `hello1` and `hello2` to the node `hello3` but we use the same requirement `input_component` for both of them:

```yaml

node_types:
  hello_type_1:
    attributes:
      attribute1:
        type: string
        default: attribute1

  hello_type_2:
    attributes:
      attribute2:
        type: string
        default: attribute2

  hello_type_3:
    properties:
      property1:
        type: string
      property2:
        type: string
    requirements:
      - input_component:
          capability: tosca.capabilities.Node
          node: tosca.nodes.SoftwareComponent
          relationship: tosca.relationships.DependsOn

topology_template:
  node_templates:
    my-workstation:
      type: tosca.nodes.Compute
      attributes:
        private_address: localhost
        public_address: localhost

    hello1:
      type: hello_type_1
      requirements:
        - host: my-workstation

    hello2:
      type: hello_type_2
      requirements:
        - host: my-workstation

    hello3:
      type: hello_type_3
      properties:
        property1: { get_attribute: [ SELF, input_component, attribute1 ] }
        property2: { get_attribute: [ SELF, input_component, attribute2 ] }
      requirements:
        - host: my-workstation
        - input_component: hello1
        - input_component: hello2

  outputs:
    output_prop1:
      description: Property output
      value: { get_property: [ hello3, property1 ] }
    output_prop2:
      description: Property output
      value: { get_property: [ hello3, property2 ] }
```

When we run this with `opera deploy` opera proceeds with the deployment without any error. And when we want to print out the outputs to check whether the connection to both nodes was successful and we were able to access their attributes, we finally get an error saying: `Targeting more than one instance via 'input_component'.` which should appear before the deployment was started.

## Steps
To fix this we would probably need to update opera's parser to search for duplicate requirements within the same node template and then throw an error.

## Current behaviour
Right now opera proceeds with the deployment when targeting more than one instance via the same requirement that can have only occurrence.

## Expected results
The orchestrator should throw an error when targeting more than one instance within the same requirement that can have only occurrence.
_________
cc @dradx, @MysteriousWolf

anzoman:
These changes correspond to issue #88 that has brought up the problem
when targeting more than one instance via the same requirement. When
the same requirement name occurred among the node type/template
requirements opera proceeded with the deployment and the error was
thrown only when running opera outputs that were accessing properties
and attributes via TOSCA functions which already contain checks for
duplicate capability and requirement names. TOSCA standard also states
that duplicate requirement names within the same TOSCA type or template
shall result in error. So now when duplicate requirement name occurs,
opera raises an exception before the deployment.

Closes #88

### DIFF
--- a/examples/attribute_mapping/service.yaml
+++ b/examples/attribute_mapping/service.yaml
@@ -31,6 +31,7 @@ node_types:
       - student:
           capability: tosca.capabilities.Root
           relationship: steampunk.test.relationships.TeacherTeachesStudent
+          occurrences: [ 0, UNBOUNDED ]
     attributes:
       student_ids:
         type: list
--- a/src/opera/parser/tosca/v_1_3/node_template.py
+++ b/src/opera/parser/tosca/v_1_3/node_template.py
@@ -87,9 +87,12 @@ class NodeTemplate(CollectorMixin, Entity):
                     node_name, name, assignment.node.data,
                 )
 
+            occurrences = definitions[name].get("occurrences")
+
             requirements.append(Requirement(
                 name, assignment.node.data,
                 relationship.get_template(relationship_name, service_ast),
+                occurrences,
             ))
 
         return requirements
--- a/src/opera/parser/tosca/v_1_3/range.py
+++ b/src/opera/parser/tosca/v_1_3/range.py
@@ -9,7 +9,7 @@ from ..base import Base
 #  1. TOSCA's range type is a bit ugly to parse because upper bound can be
 #     string (UNBOUNDED) and should be treated as infinity.
 #  2. Python's bool type subclasses int, which makes isinstance(False, int)
-#     truthy. Thsi is why we have separate checks for bool.
+#     trusty. This is why we have separate checks for bool.
 
 class Range(Base):
     @classmethod
--- a/src/opera/template/node.py
+++ b/src/opera/template/node.py
@@ -1,3 +1,5 @@
+from collections import Counter
+
 from opera.error import DataError
 from opera.instance.node import Node as Instance
 from pathlib import Path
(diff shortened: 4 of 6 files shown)


miwurster:
Pls see my comment here: https://github.com/xlab-si/xopera-opera/issues/88#issuecomment-681824110

anzoman:
Thanks for the comment @miwurster!

Since we will probably not need to do anything more here, this PR can be close or, even better, restructured to fix the problem of targeting same requirement names with `opera outputs`.

anzoman:
I also made a comment in the issue https://github.com/xlab-si/xopera-opera/issues/88#issuecomment-685366095. In the end, it seems that this PR is going in the right way and will therefore be kept and updated.

matejart (commented):

cankarm:
@alexmaslenn  can you please review this?

anzoman:
@alexmaslenn maybe you overlooked Matic's message. I would also be in favour of putting this PR forward because it's opened for almost too long now. So, I kindly ask you to make a review when you find the time.

alexmaslenn:
@anzoman sure will do

alexmaslenn (approved):
@anzoman Good job finally bringing occurrences for requirements into xOpera. Run a few tests, everything seems to work smoothly. Next step would probably be adding occurrences for capabilities, but not sure whether we really need them though

Issue #86, opened 2020-08-04
Add support for TOSCA Intrinsic functions


anzoman:
## Description
[Intrinsic functions](https://docs.oasis-open.org/tosca/TOSCA-Simple-Profile-YAML/v1.3/csprd01/TOSCA-Simple-Profile-YAML-v1.3-csprd01.html#_Toc9262317) can be useful when manipulating data within the TOSCA template.   

There are currently three functions of this kind that will need to be supported:

1. **concat**

This is used when you want to concatenate two or more string values or strings that are result of other TOSCA functions (like `get_attribute`, `get_property` or `get_input`).

```yaml
outputs:
  description: Concatenate the URL for a server from other template values
  server_url:
  value: { concat: [ 'http://',
                     get_attribute: [ server, public_address ],
                     ':',
                     get_attribute: [ server, port ] ] }
```

2. **join**

This will be used to combine an array of strings into a single string (with an optional delimiter).

```yaml
outputs:
   example1:
       # Result: prefix_1111_suffix
       value: { join: [ ["prefix", 1111, "suffix" ], "_" ] }
   example2:
       # Result: 9.12.1.10,9.12.1.20
       value: { join: [ { get_input: my_IPs }, “,” ] } 
```

3. **token**

Used on a string to parse out (tokenize) substrings separated by one or more token characters within a larger string.

```yaml
outputs:
   webserver_port:
     description: the port provided at the end of my server’s endpoint’s IP address
     value: { token: [ get_attribute: [ my_server, data_endpoint, ip_address ], ‘:’, 1 ] }
```

## Steps
To implement this we would need to update opera's TOSCA parser and probably add a similar behaviour as we have for property and attribute TOSCA functions (like `get_attribute`, `get_property` and `get_input`).

## Current behaviour
Right now xOpera does not support any of intrinsic functions

## Expected results
To be able to invoke TOSCA intrinsic functions (concat, join, token).
_________
cc @dradx

dradX:
@anzoman Thanks for the implementation of intrinsic TOSCA functions - merging this soon would be highly appreciated.

anzoman:
I'm closing this issue cause this functionality has been developed within #91 and will be available in the next opera release (and for testing purposes it is already available on the Test PyPI instance here: https://test.pypi.org/project/opera/0.6.2.dev4/).

PR #91, opened 2020-08-24
Provide support for the intrinsic TOSCA functions

anzoman:
Within this commit we are adding the support for TOSCA intrinsic
functions which can be useful when manipulating data in TOSCA
templates. The desire to provide this began with issue #86. Then we
implemented parsing and processing of all three intrinsic functions,
which are concat, join and token along with all their parameters.
There is also an example showing how to use these functions along with
opera outputs that can be found in examples/intrinsic_functions.
__________________________________________________________
cc @dradX, @MysteriousWolf

### DIFF
--- a/src/opera/template/topology.py
+++ b/src/opera/template/topology.py
@@ -68,3 +68,92 @@ class Topology:
         node_name, *rest = params
 
         return self.find_node(node_name).get_artifact(["SELF"] + rest)
+
+    def concat(self, params):
+        if not isinstance(params, list):
+            raise DataError("Concat intrinsic function parameters '{}'"
+                            " should be a list".format(params))
+
+        return self.join([params])
+
+    def join(self, params):
+        if 1 <= len(params) <= 2:
+            if not isinstance(params[0], list):
+                raise DataError("Concat or join intrinsic function parameters "
+                                "'{}' should be a list".format(params))
+
+            string_value_expressions = params[0]
+            delimiter = "" if len(params) == 1 else params[1]
+
+            if not isinstance(delimiter, str):
+                raise DataError(
+                    "Delimiter: '{}' should be a string.".format(delimiter))
+
+            values_to_join = []
+            for param in string_value_expressions:
+                if isinstance(param, dict):
+                    param_dict_keys = list(param.keys())
+                    if len(param_dict_keys) > 1:
+                        raise DataError(
+                            "Dict value expression for concat or join "
+                            "function has too many keys: '{}'".format(param))
+
+                    value_expression_name = param_dict_keys[0]
+                    try:
+                        values_to_join.append(
+                            getattr(self, value_expression_name)(
+                                param[value_expression_name]))
+                    except Exception:
+                        raise DataError(
+                            "Invalid value expression for concat or join "
+                            "function: '{}'".format(value_expression_name))
+                elif isinstance(param, str):
+                    values_to_join.append(param)
+                else:
+                    raise DataError(
+                        "Invalid value for concat or join function function: "
+                        "'{}'".format(param))
+
+            return delimiter.join(values_to_join)
+        else:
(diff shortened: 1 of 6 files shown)


src/opera/template/topology.py
```diff
@@ -68,3 +68,117 @@ def get_artifact(self, params):
...
+                        "function: '{}'".format(value_expression_name))
+
+            elif isinstance(param, str):
+                concatenated_string += param
+            else:
+                raise DataError(
+                    "Invalid value for concat function: '{}'".format(param))
+        return concatenated_string
+
+    def join(self, params):
```
alexmaslenn, line 103: If **join** function with 1 parameter is the same as **concat** I would get rid of **concat** implementation and do something like 
```python
    def concat(self, params):
        if not isinstance(params, list):
            raise DataError("Concat intrinsic function parameters for '{}'"
                            " should be a list".format(params))

        return self.join([params])
```

and for **join**
```python
delimiter = "" if len(params) == 1 else params[1]
```

And for all errors use something like 
```
for concat or join function
```

src/opera/template/topology.py
```diff
@@ -68,3 +68,117 @@ def get_artifact(self, params):
         node_name, *rest = params
 
         return self.find_node(node_name).get_artifact(["SELF"] + rest)
+
+    def concat(self, params):
+        if not isinstance(params, list):
+            raise DataError("Concat intrinsic function parameters for '{}'"
```
alexmaslenn, line 74: If user provides wrong parameter for **join** function, e.g something like 
```yaml
 join: [ [ "t", "o", "s", "c", "a", { get_input: input, get_input: input } ] ] 
```
the error would be
```
Dict value expression for concat function has too many keys: { get_input: input, get_input: input }
```
Which might be confusing as **concat** function is not used

alexmaslenn (commented):
Thanks Anze, good job! This implementation would be helpful. All test work fine, I've just some minor concerns.

src/opera/template/topology.py
anzoman: I agree, this may confuse someone. However, I don't see it as a problem since user used two elements inside the dict value expression, which as expected resulted in error. I didn't see an example in TOSCA that uses concat or join intrinsic function with dict that has multiple elements. Or have I missed something? What do you suggest here?

src/opera/template/topology.py
anzoman: Thanks, good suggestion! I've refactored this part.

src/opera/template/topology.py
alexmaslenn: Sorry if it was confusing. I just meant that user used **join** function but got error for **concat** which might be misleading, since **concat** may be not used at all. The sample is just the piece of TOSCA I used to test the error.

anzoman: Ahaa I get it now :smiley:. I have corrected this part in join so it would say: `Dict value expression for concat or join function has too many keys...`.

src/opera/template/topology.py
```diff
@@ -68,3 +68,88 @@ def get_artifact(self, params):
...
+
+    def concat(self, params):
+        if not isinstance(params, list):
+            raise DataError("Concat intrinsic function parameters '{}'"
+                            " should be a list".format(params))
+
+        return self.join([params])
+
+    def join(self, params):
+        if isinstance(params[0], list):
```
alexmaslenn, line 80: Now error becomes a little confusing - if the 1 parameter is not a `list `it would return `Too many or too less parameters`. We should probably keep checking the number of parameters

```suggestion
        if 1 <= len(params) <= 2:
            if not isinstance(params[0], list):
                raise DataError("Concat or join intrinsic function"
                                " parameters for '{}'"
                                " should be a list".format(params))
```

anzoman: Yeah, true. I've applied your suggested change.

alexmaslenn (approved):

PR #94, opened 2020-08-25
Establish a new CI routine to test the opera examples

### LINKED ISSUE #93: Move opera examples to integration test folder
## Description
Opera examples currently reside in [`/examples`](https://github.com/xlab-si/xopera-opera/tree/master/examples) folder, whereas opera runnable integration tests are located in [`/tests/integration`](https://github.com/xlab-si/xopera-opera/tree/master/tests/integration) directory. Both folders contain runnable TOSCA templates and playbooks, so there is no need to keep them separated. It would be nice if we could test also the TOSCA features that are present among examples but not in integration test which would allow us to be sure that we didn't break anything when adding new feature as opera integration tests are executed every time on push. Therefore I suggest moving opera's examples to integration tests folder.

## Steps
Implementation here should be straightforward. We would just move subfolders from `examples` to `tests/integration` folder and add `runme.sh` bash script so that every integration test could be runnable by CI/CD.

## Current behaviour
Right now opera keeps examples and integration tests separated.

## Expected results
To transfer all the examples in integration test folder.
______________
cc @sstanovnik this is what we talked about. However, we also agreed that we cannot move the examples that include any server establishing features (e.g. [OpenStack and nginx example](https://github.com/xlab-si/xopera-opera/tree/master/examples/nginx_openstack)) that you cannot easily test with CI/CD. So what should we do with these?

anzoman:
Where to begin? Tests, examples, more examples in the other repository.
Users can experience a total confusion. But we will not allow this to
happen with opera. That's why we opened #93 issue in order to clean up
the mess and move all the existing opera examples to the integration
test folder. This includes examples in this repository and also the
examples from xopera-examples repository that will not be needed
anymore after this. Since both examples and integration tests contain
runnable TOSCA templates and Ansible playbooks, there is no need to
keep them separated. It would be better to have just the integration
tests so that the features that were currently located only in examples
can be tested within the CI/CD pipelines too. However, it is not that
easy. Some examples require special environments (e.g. OpenStack,
nginx, AWS and so on) to be executed. For these cases we still have a
folder called endtoend_examples.
_______________________
The initial issue: #93

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,6 +14,7 @@ commands:
       - common_setup
       - run: make unit_test
       - run: make integration_test
+      - run: make examples_test
 
   run_integration_tests_using_released_opera_version:
     description: Test xOpera CLI commands with testing service template
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,11 @@ integration_test: $(integration_test_scenarios)
 $(integration_test_scenarios):
 	cd $(dir $@) && bash $(notdir $@) $(OPERA)
 
+
+.PHONY: examples_test
+examples_test:
+	cd examples && bash runme.sh $(OPERA)
+
 .PHONY: build
 build:
 	pipenv run python setup.py sdist bdist_wheel
--- a/examples/runme.sh
+++ b/examples/runme.sh
@@ -1,0 +1,54 @@
+#!/bin/bash
+set -euo pipefail
+
+# This is the script that is used to test the examples with the CI/CD.
+# You can also run it manually with: ./runme.sh opera
+
+# get opera executable
+opera_executable="$1"
+
+# test an example from ./artifacts
+echo "Testing an example from ./artifacts ..."
+mkdir artifacts/.opera
+$opera_executable deploy --instance-path artifacts/.opera artifacts/service.yaml
+$opera_executable outputs --instance-path artifacts/.opera
+
+# test an example from ./attribute_mapping
+echo "Testing an example from ./attribute_mapping ..."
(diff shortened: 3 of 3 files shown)


matejart (dismissed):
While I understand and support the goal of the PR to reduce the maintenance work needed on the developer, I disagree with the direction of approach. I've given my reasoning [in this comment](https://github.com/xlab-si/xopera-opera/issues/93#issuecomment-680618306).

I do not understand, what we achieve by renaming `examples` to `endtoend-examples`. Are there any other examples that are not end-to-end? If so, can we not then make `examples/endtoend` and `examples/other`? I suggest to simply keep `examples/`, which is something I can understand and quickly find.

I see that each integration test now has a `runme.sh` script, which, as far as I can tell, is the same for every test except for the csar test. This repetition introduces unnecessary complexity when it comes to maintenance. Why not use one common script and symlink to it from all the tests?

anzoman:
Merhaba! This PR has been now completely restructured. We are not moving the opera examples to the integration test folder anymore as this could could result in confusion for the users and it could bring a lot of problems that are described more in detail in #93. Rather then that, the first step towards the better life would be to introduce a new CI routine that would test the examples. This would be beneficial for both the users and the developers. The first ones would get an assurance that the TOSCA templates within the examples actually work and are deployable. The second ones would have an additional check that their code changes have not broken the deployment of the opera examples in any way.

Here is the piece from the CircleCI pipeline that shows the new CI routine (runnable with `make examples_test`):

image

sstanovnik (approved):

PR #97, opened 2020-09-03
Removal of needless checks for undeclared artifacts

anzoman:
These modifications of code will remove a bug that occurred when there
were different artifact names used in node templates as the ones that
were declared in the corresponding node types. The error also returned
invalid artifacts when there were no artifacts defined in the node
type, but were later added within the node template. Another firm fact
that supports removing these checks is the [TOSCA standard](https://docs.oasis-open.org/tosca/TOSCA-Simple-Profile-YAML/v1.3/cos01/TOSCA-Simple-Profile-YAML-v1.3-cos01.html#DEFN_ENTITY_ARTIFACT_DEF) itself, which
defines artifacts keyname within node types or templates as a map of
artifact definitions. This means that there are no artifact assignments
within node types that would need to correspond to artifact definitions
within the node templates.

### DIFF
--- a/src/opera/parser/tosca/v_1_3/collector_mixin.py
+++ b/src/opera/parser/tosca/v_1_3/collector_mixin.py
@@ -165,12 +165,6 @@ class CollectorMixin:
         definitions = typ.collect_artifact_definitions(service_ast)
         assignments = self.get("artifacts", {})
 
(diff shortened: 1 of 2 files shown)


cankarm:
LGTM!

Issue #100, opened 2020-09-22
Parallel execution does not stop if one execution branch fails.


alexmaslenn:
## Description
The idea behind parallelisation of workflow execution was that execution would stop as soon as it encounters an error. Currently execution continues if some nodes were scheduled for execution before an error was encountered.

## Steps
Deploy a service template with 2 independent nodes where an error occurs while deploying a first node, e.g
```yaml
topology_template:
  node_templates:
    my-workstation:
      type: tosca.nodes.Compute
      attributes:
        private_address: localhost
        public_address: localhost

    hello-2: # <- fail on this node
      type: hello_type
      properties:
        time: "2"
      requirements:
        - host: my-workstation     
        
    hello-3:
      type: hello_type
      properties:
        time: "1"
      requirements:
        - host: my-workstation  
```

## Current behaviour
xOpera would proceed to deploy other node.

## Expected results
xOpera should stop ASAP when encountering an error.

anzoman:
@alexmaslenn I'm closing this issue since it was fixed with #103 (it seems that GitHub does not allow fixing issues with comments that are created after the initial PR's description so I had to close this manually).

PR #103, opened 2020-09-23
Added fix for concurrent immediate cancellation

alexmaslenn:
Deployment/undeployment do not schedule nodes for execution
unless there are free workers to avoid pending operations. Also upon
error executor tries to cancel pending operations if there are any.


alexmaslenn:
Fixes #100

anzoman (approved):
Thanks @alexmaslenn! LGTM. I've also tested it locally and now the execution of all workers ends immediately when user cancels it.

Issue #104, opened 2020-09-30
Introduce ```opera info``` and ```opera set``` command


cankarm:
The ```opera info``` command would print out the current initialization of the project. In other words, ```opera init``` prepares everything to deploy, while ```opera info``` would provide the printout of:

- root blueprint file
- csar location
- inputs file name/location
- (status) initialised/deployed/error

So if the ```opera init``` fails, the user would be able to manually set some things for non-standard CSARs.

@sstanovnik, can you please review this requirement?

sstanovnik:
What would `opera set` be? A way to set these parameters? Not a good idea I think, this would be better suited for a csar authoring/publishing command. We have a PR or issue open for this I think.

Is the csar in "csar location" the extracted root? Technically incorrect, because csars are only compressed service archives. Maybe `content_root`?

Otherwise this seems fine. I'd just like to propose a schema so we don't have freeform output. My quasi-spec suggestion:

```json
{
    "entrypoint":             "string | null",
    "csar (or content_root)": "string",
    "inputs_file":            "string | null",
    "status":                 "oneof(initialised, deployed, error)"
}
```

If you want freeform as default, please also include `opera info [--format <text|json>]` so this is machine-parseable.

cankarm:
With ```opera set``` was the idea to manually set init values for development. In this case you would be able to update init values, which are then retrieved by SaaS (with info command). This would allow retrieving data of the last execution.

sstanovnik:
Info command added in #106, the decision about set postponed.

PR #106, opened 2020-10-06
Summon opera info command

anzoman:
Within these changes we are bringing a little sunshine. The main
addition is the new opera info command, which is the sixth opera CLI
command in addition to opera validate, init, deploy, undeploy and
outputs. The idea for this command originated from #104, where we
wanted to have an option to inform user about the current opera
project and would allow user to access its storage and state. This
would include printing out the path to TOSCA service template
entrypoint, extracted CSAR location, path to the storage inputs and
status/state of the deployment. The output can be formatted in YAML
or JSON. An example of the latter is shown below.

```json
{
   "service_template":  "string | null",
   "content_root":      "string | null",
   "inputs":            "string | null",
   "status":            "initialized | deployed | undeployed | interrupted | null"
}
```

This commit also includes usage of opera info command within the
integration test scripts to be sure that everything works ok.
cc @cankarm, @sstanovnik

### DIFF
--- a/src/opera/commands/info.py
+++ b/src/opera/commands/info.py
@@ -1,0 +1,99 @@
+import argparse
+import json
+from os import path
+from pathlib import Path, PurePath
+
+import yaml
+
+from opera.error import DataError, ParseError
+from opera.parser import tosca
+from opera.storage import Storage
+
+
+def add_parser(subparsers):
+    parser = subparsers.add_parser(
+        "info",
+        help="Show information about the current project"
+    )
+    parser.add_argument(
+        "--instance-path", "-p",
+        help=".opera storage folder location"
+    )
+    parser.add_argument(
+        "--format", "-f", choices=("yaml", "json"), type=str,
+        default="yaml", help="Output format",
+    )
+    parser.add_argument(
+        "--verbose", "-v", action='store_true',
+        help="Turns on verbose mode",
+    )
+    parser.set_defaults(func=_parser_callback)
+
+
+def format_outputs(info, format):
+    if format == "json":
+        return json.dumps(info, indent=2)
+    if format == "yaml":
+        return yaml.safe_dump(info, default_flow_style=False)
+
+    assert False, "BUG - an invalid format got past the parser"
+
+
+def _parser_callback(args):
+    if args.instance_path and not path.isdir(args.instance_path):
+        raise argparse.ArgumentTypeError("Directory {0} is not a valid path!"
+                                         .format(args.instance_path))
+
+    storage = Storage.create(args.instance_path)
+    try:
+        outs = info(storage)
+        print(format_outputs(outs, args.format))
+    except ParseError as e:
+        print("{}: {}".format(e.loc, e))
+        return 1
+    except DataError as e:
+        print(str(e))
+        return 1
+    return 0
+
+
+def info(storage: Storage) -> dict:
+    """
+    :raises ParseError:
+    :raises DataError:
+    """
+    info_dict = dict(
+        service_template=None,
+        content_root=None,
+        inputs=None,
+        status=None
+    )
+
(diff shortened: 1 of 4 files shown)


src/opera/commands/info.py
```diff
@@ -0,0 +1,99 @@
...
+
+from opera.error import DataError, ParseError
+from opera.parser import tosca
+from opera.storage import Storage
+
+
+def add_parser(subparsers):
+    parser = subparsers.add_parser(
+        "info",
+        help="Retrieve service template outputs"
```
sstanovnik, line 16:
```suggestion
        help="Show information about the current project."
```

Maybe not project, but workspace? Or another word more suitable? Not sure here.

src/opera/commands/info.py
```diff
@@ -0,0 +1,99 @@
...
+    parser.set_defaults(func=_parser_callback)
+
+
+def format_outputs(info, format):
+    if format == "json":
+        return json.dumps(info, indent=2)
+    if format == "yaml":
+        return yaml.safe_dump(info, default_flow_style=False)
+
+    assert False, "BUG - invalid format"
```
sstanovnik, line 39:
```suggestion
    assert False, "BUG - an invalid format got past the parser"
```

tests/integration/misc-tosca-types/runme.sh
```diff
@@ -6,9 +6,13 @@ opera_executable="$1"
 
 # perform integration test with service template
 $opera_executable validate -i inputs.yaml service-template.yaml
+$opera_executable info
 $opera_executable deploy -i inputs.yaml service-template.yaml
+$opera_executable info
 $opera_executable outputs
+$opera_executable info --format yaml
 $opera_executable undeploy
+$opera_executable info --f yaml
```
sstanovnik, line 15: This should be `-f`, right?

tests/integration/misc-tosca-types/runme.sh
```diff
@@ -6,9 +6,13 @@ opera_executable="$1"
 
 # perform integration test with service template
 $opera_executable validate -i inputs.yaml service-template.yaml
+$opera_executable info
```
sstanovnik, line 9: I'd really like it if you added at least one test for the contents, eg

```suggestion
info_out="$($opera_executable info --format json)"
test "$(echo "$info_out" | jq -r .status)" = "null"
# and other tests
```

sstanovnik (changes requested):
When can the status be null? I can't see that code path. You also have "interrupted", which doesn't show up in your spec.

Not much to change otherwise, but I'd really like the test to include output content like in the example below.

src/opera/commands/info.py
anzoman: I think project is alright.

src/opera/commands/info.py
anzoman: Refactored.

tests/integration/misc-tosca-types/runme.sh
anzoman: Yeah, I fixed this.

tests/integration/misc-tosca-types/runme.sh
anzoman: Ok, I added a few of these lines to the script.

anzoman:
@sstanovnik thanks sir for your comments. The status can be `null` at the beginning. When you run opera init, deploy or undeploy the status changes and with opera outputs and validate it stays the same. The `interrupted` status is used for example when some of instances are deployed and some are just initialized (undeployed), so the deployment/undeployment process has probably been interrupted due to some error or manually by the user. I added this to the commit description as well.

sstanovnik (approved):

sstanovnik:
Looks nice!

Issue #105, opened 2020-09-30
Error handling with multiple ```opera init``` invocations


cankarm:
Executing ```opera init``` multiple times should be prohibited. In case of different CSARs this could result in merging two incompatible CSARs together and user should be notified about that. Before new init, previous init files/folders should be deleted.

anzoman:
PR #109 will fix this.

PR #109, opened 2020-10-08
Apply fixes to opera CLI commands

anzoman:
Here we apply some changes for opera CLI commands, which are important
for better performance for the users and more readable code for the
developers. The fixes and improvements we are talking about are:

- fix imports within the code (ordering, unused imports)
- opera validate can now be used also with compressed CSARs
- use opera info state in deploy and undeploy commands
- add opera undeploy --resume/-r and --force/-f flags
- add opera init --clean/-c flag that solves the problem spotted in #105
- add integration test for all opera commands


tests/integration/cli_commands/runme.sh
```diff
@@ -0,0 +1,136 @@
...
+test "$(echo "$info_out" | jq -r .status)" = "deployed"
+
+# init service template again (with clean option and with inputs)
+$opera_executable init --clean service.yaml
+
+# test opera info status after init
+info_out="$($opera_executable info --format json)"
+test "$(echo "$info_out" | jq -r .status)" = "initialized"
+
+# deploy service template, but interrupt after 0.5 seconds
```
sstanovnik, line 52: after 1 second, right? :)

sstanovnik (changes requested):
These changes all seem sensible. But it's all in one commit, which is bad because there are _sooo_ many different things you've changed. And it's not just "minor fixes to commands", you have new functionality in there, along with new tests and formatting changes. I suggest splitting into:

* tests

src/opera/commands/init.py
```diff
@@ -38,11 +43,20 @@ def add_parser(subparsers):
...
 
+    if storage.exists("root_file"):
+        if args.clean:
+            storage.remove_all()
+            storage.create(args.instance_path)
+        else:
+            print("Looks like service template or CSAR has already been "
+                  "initialized. Use --clean/-c flag to clear the storage.")
+            return 1
+
```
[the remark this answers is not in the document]
anzoman: Done

tests/integration/cli_commands/runme.sh
anzoman: Yass

anzoman:
@sstanovnik thanks for your comments. I have made some changes. And I agree that one commit is not enough so we have 7 commits now.

matejart (commented):

sstanovnik (approved):

PR #111, opened 2020-10-13
Use all artifacts when accessing them with get_artifact function

anzoman:
These changes apply mostly to the usage of get_artifact TOSCA function.
Up until now opera used just the artifacts defined within node types
and when addressing them with `get_artifact` everything worked okay.
Fine, right? No, there's more that meets the eye. Artifacts can also be
defined right before the deployment within the node template itself. So
we needed to take those into account too, so that they can be accessed
with `get_artifact` function. Since TOSCA does not specially define
artifact assignments, there is no priority between them. This could
then result in error if user defines the artifact with the same name in
node type and node template. However, TOSCA states that when having
duplicated artifacts, the ones in node template should augment those
that were provided by its declared node type. So this still leaves us
with the idea that we need to prioritize node templates' artifacts. We
have also updated an example in `/examples/artifacts` to test both 
types of artifact definitions.

### DIFF
--- a/src/opera/parser/tosca/v_1_3/collector_mixin.py
+++ b/src/opera/parser/tosca/v_1_3/collector_mixin.py
@@ -6,7 +6,7 @@ from opera.template.capability import Capability
 class CollectorMixin:
     def collect_types(self, service_ast):
         typ = self.type.resolve_reference(service_ast)
-        return (self.type.data, ) + typ.collect_types(service_ast)
+        return (self.type.data,) + typ.collect_types(service_ast)
 
     def collect_properties(self, service_ast):
         typ = self.type.resolve_reference(service_ast)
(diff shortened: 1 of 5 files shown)


src/opera/parser/tosca/v_1_3/collector_mixin.py
```diff
@@ -165,6 +166,14 @@ def collect_artifacts(self, service_ast):
         definitions = typ.collect_artifact_definitions(service_ast)
         assignments = self.get("artifacts", {})
 
+        duplicate_interfaces = set(assignments.keys()).intersection(
+            definitions.keys())
+        if duplicate_interfaces:
+            for duplicate in duplicate_interfaces:
+                definitions.pop(duplicate)
+
+        definitions.update(assignments)
```
cankarm, line 175: If I understand correctly this is the main change here (most of the others are updating the style)?

Seems OK by me.

anzoman: Yes, here I extended the usage of artifacts so that ones from node templates can be also used. To be sure this works I added some content to the examples so that you can run `opera deploy /examples/artifacts/service.yaml` and the `opera outputs` to see that both file contents were included.

This should be the result:
image

cankarm: I saw the example, but at first glance, I understood that we want to overload the definition (and therefore re-define the first one). Therefore I thought that the example does not reflect the usage what we want to achieve.  Thanks for your clarification @anzoman !

cankarm (approved):
It seems OK. This can be merged.

PR #113, opened 2020-10-15
Use node name references with intrinsic functions

### LINKED ISSUE #112: Concat does not work with nodes own values
When using `concat` intrinsic function (`with opera==0.6.2.dev3`) it fails when trying to concatenate node's own values.

This works fine:
`attribute: { get_attribute: [ SELF, property ]}`

this also works fine:
`attribute: { concat: [ 'This is property: ', get_attribute: [ other_node, property ]]}`

but this fails with "Invalid value expression for concat or join function: 'get_attribute'":
`attribute: { concat: [ 'This is property: ', get_attribute: [ SELF, property ]]}`

Here is full test file:
```yaml

node_types:
  hello_type:
    attributes:
      attribute:
        type: string
    properties:
      property:
        type: string


topology_template:

  node_templates:
    my-workstation:
      type: tosca.nodes.Compute
      attributes:
        private_address: localhost
        public_address: localhost

    hello:
      type: hello_type
      attributes:
        attribute: { get_attribute: [ SELF, property ]}
        # this doesn't work
        # attribute: { concat: [ 'This is property: ', get_attribute: [ SELF, property ]]}
      properties:
        property: "ThatIsProperty"
      requirements:
        - host: my-workstation

  outputs:
    attribute:
      description: Value of attribute
      value: { get_attribute: [ hello, attribute ]}
```

anzoman:
In this commit we apply some refactorings to TOSCA instrinsic functions
based on the problems spotted in #112. There it was discovered that
SELF keyword in TOSCA functions do not work when using them inside
intrinsic functions. Or in other words, we are unable to get node's own
attributes and properties from intrinsic functions used within this
node. The problem originates from the part where we have forgotten that
we the intrinsic functions will not be used only in template's outputs,
but also throughout the whole TOSCA template. Therefore this was
calling for a change where we made sure that node names and their
references are used when calling other TOSCA functions in intrinsic
functions. This does mostly concern concat and join functions as those
two are allowed to have other function calls (e.g. get_attribute,
get_property, get_input) within their list of string value expressions.
Token function is not included because it can operate on strings only.
______

cc @mihaTrajbaric, @dradx

Closes #112.

### DIFF
--- a/src/opera/template/node.py
+++ b/src/opera/template/node.py
@@ -197,10 +197,10 @@ class Node:
             raise DataError("Cannot find artifact '{}'.".format(prop))
 
     def concat(self, params):
-        return self.topology.concat(params)
+        return self.topology.concat(params, self)
 
     def join(self, params):
-        return self.topology.join(params)
+        return self.topology.join(params, self)
 
     def token(self, params):
         return self.topology.token(params)
--- a/src/opera/template/topology.py
+++ b/src/opera/template/topology.py
@@ -69,14 +69,14 @@ class Topology:
 
         return self.find_node(node_name).get_artifact(["SELF"] + rest)
 
-    def concat(self, params):
+    def concat(self, params, node=None):
         if not isinstance(params, list):
             raise DataError("Concat intrinsic function parameters '{}'"
                             " should be a list".format(params))
 
-        return self.join([params])
+        return self.join([params], node)
 
-    def join(self, params):
+    def join(self, params, node=None):
         if 1 <= len(params) <= 2:
             if not isinstance(params[0], list):
                 raise DataError("Concat or join intrinsic function parameters "
@@ -100,8 +100,9 @@ class Topology:
 
                     value_expression_name = param_dict_keys[0]
                     try:
+                        entity = node if node else self
                         values_to_join.append(
-                            getattr(self, value_expression_name)(
+                            getattr(entity, value_expression_name)(
                                 param[value_expression_name]))
                     except Exception:
                         raise DataError(
--- a/examples/intrinsic_functions/service.yaml
+++ b/examples/intrinsic_functions/service.yaml
@@ -11,13 +11,30 @@ node_types:
         type: string
 
   hello_type_2:
+    derived_from: tosca.nodes.SoftwareComponent
     attributes:
       attribute2:
         type: string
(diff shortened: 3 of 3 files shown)


matejart (commented):

src/opera/template/topology.py
```diff
@@ -100,8 +100,10 @@ def join(self, params):
 
                     value_expression_name = param_dict_keys[0]
                     try:
+                        entity_name = self.find_node(
```
alexmaslenn, line 103: This is a little confusing - we are actually dealing here not with entity name but entity itself 
```suggestion
                        entity = self.find_node(
```

src/opera/template/topology.py
```diff
@@ -100,8 +100,10 @@ def join(self, params):
...
+                            node_name) if node_name else self
                         values_to_join.append(
-                            getattr(self, value_expression_name)(
+                            getattr(entity_name, value_expression_name)(
```
alexmaslenn, line 106:
```suggestion
                            getattr(entity, value_expression_name)(
```

alexmaslenn (changes requested):

src/opera/template/node.py
```diff
@@ -197,10 +197,10 @@ def get_artifact(self, params):
             raise DataError("Cannot find artifact '{}'.".format(prop))
 
     def concat(self, params):
-        return self.topology.concat(params)
+        return self.topology.concat(params, self.name)
```
alexmaslenn, line 200: What is the point of providing name an then searching by name later? Why not provide node reference directly like
```suggestion
        return self.topology.concat(params, self)
```

alexmaslenn (changes requested):

src/opera/template/topology.py
anzoman: Yes, true. Here, we are actually getting an entity but I named it ambiguously, so I have changed it to just `entity`.

src/opera/template/topology.py
anzoman: Fixed.

src/opera/template/node.py
anzoman: There's no point at all, so this has been optimized.

alexmaslenn (approved):

Issue #89, opened 2020-08-10
Consider adding an option to disable Ansible host key checking


anzoman:
## Description
Ansible has host key checking enabled by default which also applies when running Ansible playbooks as TOSCA actuators with opera. But sometimes this can bring some problems, for instance with use cases when a new virtual machine gets spinned up for every new deployment. This means that user may connect to the VM for the first time when he initiates the deployment so he may be prompted to type in `[yes/no]` to continue connecting to the VM. This can be an issue for example when opera deployment are running in CI/CD pipelines and simple if user doesn't want to type in anything and just wants to continue with the deployment. In this case it might be helpful if he could disable that prompting by setting Ansible host key checking to false. But again this could not be done globally because checking host keys protects users against server spoofing and MITM attacks. So in cases when user understands the possible risks, it would be better to have an environment variable to control that behaviour (like `OPERA_SSH_HOST_KEY_CHECKING`) which can be set to false.

## Steps
To add this feature we need to update opera's Ansible executor to modify `ansible.cfg` file. This can be done by adding similar code snippet as the one that is below:

```python
opera_ssh_host_key_checking = os.environ.get("OPERA_SSH_HOST_KEY_CHECKING")
if opera_ssh_host_key_checking is not None:
    check = str(opera_ssh_host_key_checking).lower().strip()
    if check[:1] == 'f' or check[:1] == "false":
        with open("{}/ansible.cfg".format(dir_path), "w") as fd:
            fd.write("[defaults]\n")
            fd.write("host_key_checking = False\n")
```

Another possible way would be to set env var `ANSIBLE_HOST_KEY_CHECKING` to False.

## Current behaviour
Right now opera doesn't have the ability to disable Ansible host key checking (opera currently just sets `StrictHostKeyChecking` to no, which means that opera will automatically add new host keys to the user known hosts files).

## Expected results
The orchestrator could allow disabling host key checking.

duelle:
As we need to spin up instances without manual interaction, I'm very much in favor of this!

anzoman:
We are exploring this issue. Currently, it seems that this will be the solution: https://docs.ansible.com/ansible/latest/user_guide/connection_details.html#managing-host-key-checking.

sstanovnik:
Resolved in #114.

PR #114, opened 2020-10-15
Add an option to disable Ansible host key checking

anzoman:
The issue #89 described the story behind Ansible host key checking and
the fact that there are situations when host checks need to be set to
false. Hitherto, opera disabled strict host key checking within the
Ansible inventory. The changes here provide the ability to disable host
key checking globally (within ansible.cfg) which can be achieved by
using `OPERA_SSH_HOST_KEY_CHECKING` environment variable.


cankarm (approved):
LGTM

PR #116, opened 2020-10-19
Fix policies serialization and refactor examples

### LINKED ISSUE #115: Policies in service templates should be serialized as a list
## Description
This issue describe the inconsistency spotted in opera's TOSCA YAML v1.3 parser. [Section 3.9 of TOSCA Simple Profile in YAML v1.3](https://docs.oasis-open.org/tosca/TOSCA-Simple-Profile-YAML/v1.3/os/TOSCA-Simple-Profile-YAML-v1.3-os.html#_Toc26969450) states that `policies` keyname within the `topology_template` should be represented as an optional list of TOSCA policy definitions that apply to the topology template.

So, the syntax for TOSCA policies here should look like this:

```yaml
topology_template:
    node_templates:
    # left off for brevity

    policies:
      - policy1:
          type: tosca.policies.placement
          # left off for brevity
      - policy2:
          type: tosca.policies.placement
          # left off for brevity
      - policy3:
          type: tosca.policies.placement
          # left off for brevity
```

When we try to use that syntax and deploy it with opera, we get an error (`Expected map.`), saying that opera expected a YAML dictionary/map for the `policies` keyname. And by looking into opera's parser we soon realize that policies are not serialized as a list, but rather as a map which is in contradiction with the TOSCA standard and it needs to be fixed. 

## Steps
To fix the issue in the TOSCA parser, we should modify the content in `src/opera/parser/tosca/v_1_3/topology_template.py` and also fix any tests or examples that use the wrong synax. 

## Current behaviour
Right now opera treats policies in `topology_template` as YAML maps.

## Expected results
Opera's TOSCA parser should serialize policies as YAML lists.

anzoman:
These significant changes are directly connected to the issue #115. To
sum things up, here we changed the parsing of the TOSCA policies for
the topology_template section. Before this commit, policies were parsed
and represented as a map/dict and after looking at the TOSCA standard,
we soon have found out that we should have used a list. So the TOSCA
parser within the orchestrator was refactored to the point that it
serializes policies as a list. By making this change we also had to
refactor some examples and integration tests.

There is also an additional commit for updating the examples.
________________

Fixes #115.

Connected to https://github.com/radon-h2020/radon-particles/pull/62

### DIFF
--- a/examples/attribute_mapping/service.yaml
+++ b/examples/attribute_mapping/service.yaml
@@ -20,7 +20,7 @@ node_types:
         operations:
           create:
             inputs:
-              id: { default: { get_property: [ SELF, enrolment_number ] } }
+              id: { default: { get_property: [ SELF, enrolment_number ] }, type: integer }
             outputs:
               student_id: [ SELF, student_id ]
             implementation: playbooks/create-student.yaml
@@ -51,8 +51,10 @@ relationship_types:
             inputs:
               student_id:
                 default: { get_attribute: [ TARGET, student_id ] }
+                type: string
               student_ids:
                 default: { get_attribute: [ SOURCE, student_ids ] }
+                type: list
             outputs:
               new_list: [ SOURCE, student_ids ]
             implementation: playbooks/teacher-teaches-student--preconfigure-source.yaml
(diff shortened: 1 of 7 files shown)


matejart (commented):

miwurster:
Any update on this?

anzoman:
@sstanovnik and @cankarm please review this PR so that we will resolve this bug ASAP.

cankarm (approved):
There are only simple changes, I approve this one.

Issue #107, opened 2020-10-08
Opera init CLI command may be unncecessary


anzoman:
## Description
This issue is meant for the discussion about opera init command which might be uneeded.

The latest version [`0.6.1`](https://github.com/xlab-si/xopera-opera/releases/tag/0.6.1) of `opera` TOSCA orchestrator provides the following CLI commands:

```console
usage: opera [-h] {deploy,init,outputs,undeploy,validate} ...

opera orchestrator

positional arguments:
  {deploy,init,outputs,undeploy,validate}
    deploy              Deploy service template from CSAR
    init                Initialize the deployment environment for the service template or CSAR
    outputs             Retrieve service template outputs
    undeploy            Undeploy service template
    validate            Validate service template from CSAR

optional arguments:
  -h, --help            show this help message and exit
```

When user wants to deploy a (compressed) CSAR or a a logical way to do it with opera would feature using the commands below.

```console
# initialize the CSAR (from the prepared compressed CSAR file, also provide inputs)
opera init --inputs inputs.yaml test.csar

# deploy the initialized CSAR
opera deploy
```

This is all ok. Opera init checks the CSAR structure and prepares `root_file` and `input` files in opera storage and deploy initiates the deployment.

However, when you deploy a service template you can use init or you can deploy directly from the service template using `opera deploy service.yaml`. So init here is a totally unnecessary step. The same could apply for the zipped CSARs. Basically with init opera doesn't do anything special. It just checks the CSAR structure, which is done by a separated CSAR class that can be imported anywhere. So I thought why would we complicate our lives? We could move the CSAR structure check to deploy CLI command and have opera deploy for everything. Thereby user wouldn't need to do one additional step with init and he would just use `opera deploy my.csar` to deploy a compressed CSAR. It has also been evident that many users were confused with init because they completely forgot it when deploying their pre-packed CSARs, because they were used to do just opera deploy with all their extracted CSARs and TOSCA service templates.

I can sense that someone might think that opera init can be good when we want to distinguish compressed and extracted CSARs. I belive that init can be useful if we had some additional options for CSARs that cannot be done before/at the start of the deployment and would consume a lot of time, but currently I don't see that coming. One additional good point to throw init away would be the `opera validate` command we have which could and will be redesigned to validate not just TOSCA service template, but also compressed CSARs. So this will also be a place, where we will use our CSAR validator class and it would then be completely unnecessary to have this in init. 

## Steps
[...]

PR #123, opened 2020-11-02
Mark opera init as deprecated and enable compressed CSAR deployment with opera deploy

anzoman:
These changes correspond to the debate within the #107 issue. There we
have agreed that opera init command is unneeded at the moment as it
does not do anything different from opera deploy (apart from just
skipping the last deployment step). We can achieve the same behavior
with opera deploy by adding the compressed CSAR initialization logic.
And that's exactly what we did. Users are now able to deploy TOSCA
service templates, uncompressed CSARs or compressed CSARs with opera
deploy. We thought about the possibility that some users would still
initialize the service template/CSAR first and the deploy, so we kept
the option to run just opera deploy after the initialization.

As opera init command currently has no benefits, we are also applying the
changes that will warn the users that opera init CLI command is deprecated 
and that he can just use opera deploy instead.

### DIFF
--- a/src/opera/commands/deploy.py
+++ b/src/opera/commands/deploy.py
@@ -4,11 +4,13 @@ import yaml
 
 from os import path
 from pathlib import Path, PurePath
+from zipfile import ZipFile, is_zipfile
 
 from opera.commands.info import info
 from opera.utils import prompt_yes_no_question
 from opera.error import DataError, ParseError
 from opera.parser import tosca
+from opera.parser.tosca.csar import CloudServiceArchive
 from opera.storage import Storage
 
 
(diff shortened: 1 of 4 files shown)


sstanovnik (changes requested):
Can you add an integration test that deploys a zipped CSAR? If possible, something that is a duplicated of an existing, complex test case, with the only difference being that you deploy a zipped thing instead of a yaml. Don't duplicate the test cases though, because they'll have to be the exact same contents and tests for this to be useful.

anzoman:
@sstanovnik thanks for the review. I've added some integration tests cases to the existing integration tests so that direct CSAR deployment can be tested.

sstanovnik (approved):

PR #126, opened 2020-11-09
Mention that orchestration inputs can be also passed as JSON

### LINKED ISSUE #122: Allow passing of opera inputs in other formats (like JSON)
## Description
The orchestrator allows passing inputs while deploying TOSCA templates/CSARS. These inputs are passed as a YAML file containing key: value pairs for TOSCA inputs. The usual process features passing inputs at the beginning of the deployment like this: `opera deploy --inputs inputs.yaml service.yaml`.  It is cool that we have inputs in the similar format as TOSCA templates. However, there are cases when users would rather pass input file in some other format than YAML because their tools or use cases are tied to this other format. Possible additional format would definitely be JSON. JSON files are probably used widely than YAML files. The current version of xOperaSaaS component also receives inputs as a JSON an then transforms them into YAML. With this update of format this won't be necessary any longer. Another possible reason that supports adding JSON format for inputs is that we already have this implemented within opera outputs, where user can call use a special flag for this `-f/--format json/yaml`.  We can do this similarly with inputs (using `--inputs-format` flag) like this:

```console
# --inputs-format switch would receive yaml/json values and when not used it would default to yaml
# --inputs-format switch could be used only when --inputs switch is used
opera deploy --inputs-format json --inputs inputs.yaml service.yaml
```

So, the default behaviour would still be the same as it is now, but with this change the users of opera would not be obliged to use 
YAML inputs format and would not have to adjust their tools to much to be compatible with opera.

## Steps
We would need to modify the files in `/src/opera/commands/deploy.py` (+ files in `/src/opera/commands/validate.py` to support this also for opera TOSCA validation). 

## Current behaviour
Opera now receives inputs only in the YAML format.

## Expected results
To have more available input file formats (JSON)

anzoman:
Inputs can be very important for the orchestration process as they
often include various secrets and variables that are used to
authenticate and to deploy applications. Up until now inputs were
passed to the orchestrator as a YAML file with key-value pairs.
To make our TOSCA orchestrator more flexible we can extend inputs
format (this was originally explained in the issue #122). So, with
these changes we are adding the possibility to supply inputs also
in JSON format. This was not very hard because opera already stores
inputs as a JSON file in its storage folder (.opera by default) and
therefore we only had to modify the things for certain CLI commands.
This is the part where --inputs-format CLI option has been brought to
life. It can be used to specify the desired YAML or JSON inputs format
(YAML is used by default) for validate, init or deploy CLI command.

The new option `--inputs-format` can be used like this:

```console
# validate TOSCA service template with YAML inputs
# --inputs-format is not used and defaults to yaml
(.venv) anzoman@ubuntu:~/Desktop/xopera-opera/templates/misc-tosca-types$ opera validate -i inputs.yaml service-template.yaml
Validating service template ...
Done.

# validate TOSCA service template with JSON inputs
(.venv) anzoman@ubuntu:~/Desktop/xopera-opera/templates/misc-tosca-types$ opera validate -i inputs.json --inputs-format json service-template.yaml
Validating service template ...
Done.

# initialize TOSCA service template with YAML inputs
(.venv) anzoman@ubuntu:~/Desktop/xopera-opera/templates/misc-tosca-types$ opera init -i inputs.yaml --inputs-format yaml service-template.yaml
Service template was initialized

# deploy the initialize TOSCA service template and override YAML inputs with JSON inputs
(.venv) anzoman@ubuntu:~/Desktop/xopera-opera/templates/misc-tosca-types$ opera deploy -i inputs.json --inputs-format json
[Worker_0]   Deploying my-workstation1_0
...
```
___________
Closes #122

### DIFF
--- a/src/opera/commands/deploy.py
+++ b/src/opera/commands/deploy.py
@@ -25,7 +25,7 @@ def add_parser(subparsers):
     )
     parser.add_argument(
         "--inputs", "-i", type=argparse.FileType("r"),
-        help="Optional: YAML file with inputs to "
+        help="YAML or JSON file with inputs to "
              "override the inputs supplied in init",
     )
     parser.add_argument(
--- a/src/opera/commands/init.py
+++ b/src/opera/commands/init.py
@@ -24,7 +24,7 @@ def add_parser(subparsers):
     )
     parser.add_argument(
         "--inputs", "-i", type=argparse.FileType("r"),
-        help="YAML file with inputs",
+        help="YAML or JSON file with inputs",
     )
     parser.add_argument(
         "--clean", "-c", action='store_true',
@@ -50,7 +50,6 @@ def _parser_callback(args):
                                          .format(args.instance_path))
 
     storage = Storage.create(args.instance_path)
-
     try:
         inputs = yaml.safe_load(args.inputs) if args.inputs else {}
     except Exception as e:
--- a/src/opera/commands/validate.py
+++ b/src/opera/commands/validate.py
@@ -18,7 +18,7 @@ def add_parser(subparsers):
     )
     parser.add_argument(
         "--inputs", "-i", type=argparse.FileType("r"),
-        help="YAML file with inputs",
+        help="YAML or JSON file with inputs",
     )
     parser.add_argument(
         "--verbose", "-v", action='store_true',
--- a/tests/integration/cli_commands/inputs.json
+++ b/tests/integration/cli_commands/inputs.json
@@ -1,0 +1,4 @@
+{
+  "test_attribute_input": 10,
+  "test_property_input": true
+}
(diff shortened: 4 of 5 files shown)


sstanovnik (commented):
Is this change even necessary? JSON is a subset of YAML, so every JSON file is also a valid YAML file. You shouldn't need to use a different parser.

Do JSON files already work without this change? If so, you could just document that the inputs parameter accepts both JSON and YAML and then just treat both the same.

anzoman:
@sstanovnik you are right! We already have that functionality but it was not mentioned anywhere. I've documented that within the commands helpers and have added some integration tests cases to be sure that inputs can be passed as JSON files.

sstanovnik (approved):
Looks good!

PR #128, opened 2020-11-14
corrected operation name to match interface

Legion2:
There is no operation called `configure_target` on the `tosca.interfaces.relationship.Configure` interface type. It should be `pre_configure_target` or `post_configure_target`.

### DIFF
--- a/tests/integration/misc-tosca-types/modules/relationship_types/test/test.yaml
+++ b/tests/integration/misc-tosca-types/modules/relationship_types/test/test.yaml
@@ -11,7 +11,7 @@ relationship_types:
     interfaces:
       Configure:
         operations:
-          configure_target:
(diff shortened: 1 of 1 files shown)


anzoman (approved):
Hi, thanks a lot for spotting this inconsistency with TOSCA Configure interface operations! We wanted to correct this in the existing PR here: https://github.com/xlab-si/xopera-opera/pull/109/files#diff-f5b735216c8c9821d29859fa2368c1237b5efa2760a4159a9e9699f1fc0e7c12L14. But since you created a PR just for that issue we can go on an merge this PR right away. 

Thanks again for your first contribution to opera. And if you find any other things that can be improved or changed you can open an issue or supply another PR 😃.

PR #132, opened 2020-12-07
Make TOSCA properties required by default

### LINKED ISSUE #131: Properties should be required by default
According to TOSCA YAML 1.3:

This is not enforced and also in the integration tests there are properties which have no value set even though they are required by default.

For example in the `tests/integration/misc-tosca-types/service-template.yaml` the `noimpl` node derived from `daily_test.nodes.noimpl` which defines two properties which don't specify the `required` keyname. So they SHALL be considered required. But the noimpl node does not assign values to the properties. This should be considered an error.

https://github.com/xlab-si/xopera-opera/blob/b987e566d1d42f1ed6d3b4d6b746aad5d8839632/tests/integration/misc-tosca-types/service-template.yaml#L68-L72

https://github.com/xlab-si/xopera-opera/blob/b987e566d1d42f1ed6d3b4d6b746aad5d8839632/tests/integration/misc-tosca-types/modules/node_types/noimpl/noimpl.yaml#L4-L12

anzoman:
Here we are addressing the issue that has been originally spotted in
#131. There is an additional requirement for property (and parameter
definitions) that has to be satisfied in order to stick with the
TOSCA Simple Profile in YAML version 1.3. Every defined property
should be considered required by default unless its required keyname is
explicitly set to false. With these changes we make sure that the users
will be warned if some required property is missing its value and in
these cases the deployment process will fail. According to that we also
needed to update some of TOSCA templates among the examples and
integration tests as there were required properties that were not used.

Fixes #131.

### DIFF
--- a/examples/intrinsic_functions/service.yaml
+++ b/examples/intrinsic_functions/service.yaml
@@ -35,6 +35,7 @@ node_types:
         default: { concat: [ 'Property: ', get_attribute: [ SELF, attribute3_2 ] ] }
       property3_3:
         type: string
+        required: false
 
 topology_template:
   inputs:
--- a/src/opera/parser/tosca/v_1_3/collector_mixin.py
+++ b/src/opera/parser/tosca/v_1_3/collector_mixin.py
@@ -19,6 +19,22 @@ class CollectorMixin:
                 ", ".join(undeclared_props),
             ), self.loc)
 
+        for key, prop_definition in definitions.items():
+            prop_required = prop_definition.get("required", None)
+            prop_has_default = prop_definition.get("default", None)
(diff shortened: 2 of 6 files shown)


sstanovnik (changes requested):
This looks good, but can you add a simple integration test that expects an error when you don't specify a required property?

anzoman:
@sstanovnik merci beaucoup for going through this PR. I think that we don't need an integration tests for that as this can be tested more easily so I have just added some simple unit tests.

sstanovnik (approved):

sstanovnik:
lgtm!

PR #142, opened 2020-12-18
Introduce opera tab completion

### LINKED ISSUE #56: Enable tab completion for opera CLI arguments
## Description
xOpera orchestrator uses [`argparse`](https://pypi.org/project/argparse/) Python library for parsing the command line arguments. This package encapsulates the option to use help for opera's commands and its subcommands which is good for the users. But apart from that it would be even better to be able to use tab completion for all of those commands and arguments as this is commonly used within prominent, prosperous and modern console clients. Thus I suggest to use [`argcomplete`](https://pypi.org/project/argcomplete/) Python package in order to implement the behaviour for optional tab-completion for the CLI.

## Steps

To implement this we would probably need to:

- use the`argcomplete` pip package and add it to `setup.cfg` and `Pipfile`
- import the package within the `xopera-opera/src/opera/cli.py` file and use it for the main parser and for the subparsers
- maybe we'll need to use the argcomplete package in the other Python files in `xopera-opera/src/opera/commands`

## Current behaviour
Currently xOpera does not offer tab completion for any of the commands or argument options. 

## Expected results
To be able to perform tab completion for the opera's commands, subcommands and argument CLI options.

anzoman:
This PR does the following:

- refactor comment in setup.py 
- add a missing verbose flag to opera package CLI command 
- implement opera tab completion

Shell completion is the most important here.

It seems that [argcomplete](https://github.com/kislyuk/argcomplete) Python package might not be the
best solution for subparsers and also has other drawbacks.
We are able to use [shtab](https://github.com/iterative/shtab) in our code to generate a shell completion
script. And we don't even need a separate command to do that since
shtab supports defining a global optional argument that will print
out the completion script for the main parser. We used that and
defined `--shell-completion/-s` flag which receives a shell type to
generate completion for. Shtab currenty supports bash and zsh so those
are the options. So, after running `opera -s bash|zsh` the generated
tab completion script will be printed out. To activate it you must
source the contents which can be done with `eval "$(opera -s bash)"`
or you can save it to a file and then source it. The procedure will
be fully explained in the opera documentation.

Closes #56.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
 pyyaml >= 3.10  # MIT
 ansible >= 2.8  # GPLv3
+shtab >= 1.3.3  # Apache 2.0
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,6 +34,7 @@ setup_requires =
 install_requires =
   ansible >= 2.8
   pyyaml >= 3.10
+  shtab >= 1.3.3
 
 [options.packages.find]
 where = src
--- a/src/opera/cli.py
+++ b/src/opera/cli.py
@@ -1,6 +1,7 @@
 import argparse
 import inspect
 import sys
+import shtab
 
 from opera import commands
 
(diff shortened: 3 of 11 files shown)


sstanovnik (approved):
This looks okay! Just rebase to fix conflicts, then we're good to go!

anzoman:
@sstanovnik rebase done.

sstanovnik (approved):

PR #144, opened 2020-12-18
fixed typo in Standard interface operation name

Legion2:


anzoman (approved):
@Legion2 thanks for spotting this, you have a sharp eye!

Legion2:
> @Legion2 thanks for spotting this, you have a sharp eye!

no it's my tosca yaml parser which does a strict validation

PR #153, opened 2020-12-31
Specialise directory and file based csars in classes

sstanovnik:
This restructures the CSAR class a bit so that it's easier to treat file- and directory-based CSARs in the same way. I also added separate meta classes to distinguish between CSAR and service template metadata.

### DIFF
--- a/src/opera/parser/tosca/csar.py
+++ b/src/opera/parser/tosca/csar.py
@@ -1,31 +1,313 @@
 import shutil
-import yaml
-
-from pathlib import Path
-from tempfile import TemporaryDirectory, mktemp
+import traceback
+from abc import ABC, abstractmethod
+from pathlib import Path, PurePath
+from tempfile import mktemp
+from typing import List, IO, Optional
 from zipfile import ZipFile
 
-from opera.error import ParseError
+import yaml
+
+from opera.error import ParseError, OperaError
+from opera.utils import determine_archive_format
 
 
-class CloudServiceArchive:
-    def __init__(self, csar_name):
-        self._csar_name = csar_name
-        self._tosca_meta = None
-        self._root_yaml_template = None
-        self._metadata = None
+class CsarMeta:
+    SUPPORTED_META_FILE_VERSIONS = {"1.1"}
+    SUPPORTED_CSAR_VERSIONS = {"1.1"}
+    METADATA_PATH = "TOSCA-Metadata/TOSCA.meta"
+
+    KEY_NAME = "Name"
+    KEY_CONTENT_TYPE = "Content-Type"
+    KEY_TOSCA_META_FILE_VERSION = "TOSCA-Meta-File-Version"
+    KEY_CSAR_VERSION = "CSAR-Version"
+    KEY_CREATED_BY = "Created-By"
+    KEY_ENTRY_DEFINITIONS = "Entry-Definitions"
+
+    REQUIRED_KEYS = {KEY_TOSCA_META_FILE_VERSION, KEY_CSAR_VERSION,
+                     KEY_CREATED_BY}
+
+    def __init__(self, name: str, content_type: str,
+                 tosca_meta_file_version: str,
+                 csar_version: str, created_by: str, entry_definitions: str):
+        self.name = name
+        self.content_type = content_type
+        self.tosca_meta_file_version = tosca_meta_file_version
+        self.csar_version = csar_version
+        self.created_by = created_by
+        self.entry_definitions = entry_definitions
+
+    @classmethod
+    def parse(cls, contents: str) -> "CsarMeta":
+        # FIXME: the meta format is not yaml, but it's close enough to parse a
+        #        simple file
+        parsed = yaml.safe_load(contents)
+
+        for req in CsarMeta.REQUIRED_KEYS:
+            if req not in parsed:
+                raise ParseError("Missing required meta entry: {}".format(req),
+                                 CsarMeta.METADATA_PATH)
+
+        csar_version = str(parsed[CsarMeta.KEY_CSAR_VERSION])
+        if csar_version not in CsarMeta.SUPPORTED_CSAR_VERSIONS:
+            raise ParseError(
+                "{} {} is not supported. Supported versions: {}\".".format(
+                    CsarMeta.KEY_CSAR_VERSION,
+                    csar_version,
+                    CsarMeta.SUPPORTED_CSAR_VERSIONS
+                ), CsarMeta.METADATA_PATH)
+
+        tmf_version = str(parsed[CsarMeta.KEY_TOSCA_META_FILE_VERSION])
+        if tmf_version not in CsarMeta.SUPPORTED_META_FILE_VERSIONS:
+            raise ParseError(
+                "{} {} is not supported. Supported versions: {}\".".format(
+                    CsarMeta.KEY_TOSCA_META_FILE_VERSION,
+                    tmf_version,
+                    CsarMeta.SUPPORTED_META_FILE_VERSIONS
+                ), CsarMeta.METADATA_PATH)
+
+        result = CsarMeta(
+            name=parsed.get(CsarMeta.KEY_NAME),
+            content_type=parsed.get(CsarMeta.KEY_CONTENT_TYPE),
+            tosca_meta_file_version=tmf_version,
+            csar_version=csar_version,
+            created_by=parsed[CsarMeta.KEY_CREATED_BY],
+            entry_definitions=parsed.get(CsarMeta.KEY_ENTRY_DEFINITIONS),
+        )
+
+        return result
+
+    def render(self) -> str:
+        ordered_fields = [
+            (CsarMeta.KEY_NAME, self.name),
+            (CsarMeta.KEY_CONTENT_TYPE, self.content_type),
+            (
+                CsarMeta.KEY_TOSCA_META_FILE_VERSION,
+                self.tosca_meta_file_version),
+            (CsarMeta.KEY_CSAR_VERSION, self.csar_version),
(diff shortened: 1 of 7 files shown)


src/opera/parser/tosca/csar.py
```diff
@@ -1,31 +1,286 @@
...
+    METADATA_PATH = "TOSCA-Metadata/TOSCA.meta"
+
+    KEY_NAME = "Name"
+    KEY_CONTENT_TYPE = "Content-Type"
+    KEY_TOSCA_META_FILE_VERSION = "TOSCA-Meta-File-Version"
+    KEY_CSAR_VERSION = "CSAR-Version"
+    KEY_CREATED_BY = "Created-By"
+    KEY_ENTRY_DEFINITIONS = "Entry-Definitions"
+
+    REQUIRED_KEYS = {KEY_TOSCA_META_FILE_VERSION, KEY_CSAR_VERSION, KEY_CREATED_BY}
```
anzoman, line 26: IDE says: `PEP 8: E501 line too long (83 > 79 characters).` Sir, please fix all the lines that are too long to be consistent with other opera Python files.

src/opera/parser/tosca/csar.py
```diff
@@ -79,16 +326,16 @@ def package_csar(self, output, service_template=None, csar_format="zip"):
...
                 content = """
-                          TOSCA-Meta-File-Version: 1.1
-                          CSAR-Version: 1.1
-                          Created-By: xOpera TOSCA orchestrator
-                          Entry-Definitions: {}
-                          """.format(service_template)
+TOSCA-Meta-File-Version: 1.1
+CSAR-Version: 1.1
+Created-By: xOpera TOSCA orchestrator
+Entry-Definitions: {}
```
anzoman, line 337: Wow, are these lines left-drunk? Please reformat.

src/opera/parser/tosca/csar.py
```diff
@@ -1,31 +1,286 @@
...
+from typing import List, IO, Optional
+from zipfile import ZipFile
+
 import yaml
 
-from pathlib import Path
-from tempfile import TemporaryDirectory, mktemp
-from zipfile import ZipFile
+from opera.error import ParseError, OperaError
+from opera.utils import determine_archive_format
```
anzoman, line 11: These imports should be split into three sections - import ___, from ___ import ___ , from opera.___ import ___.

So, this should be like this:

```python
import yaml
import shutil

from abc import ABC, abstractmethod
from pathlib import Path, PurePath
from tempfile import mktemp
from typing import List, IO, Optional
from zipfile import ZipFile

from opera.error import ParseError, OperaError
from opera.utils import determine_archive_format
```

src/opera/parser/tosca/csar.py
```diff
@@ -1,31 +1,286 @@
...
+        self.name = name
+        self.content_type = content_type
+        self.tosca_meta_file_version = tosca_meta_file_version
+        self.csar_version = csar_version
+        self.created_by = created_by
+        self.entry_definitions = entry_definitions
+
+    @classmethod
+    def parse(cls, contents: str) -> "CsarMeta":
+        # FIXME: the meta format is not yaml, but it's close enough to parse a simple file
```
anzoman, line 39: How can this be fixed in the future?

src/opera/parser/tosca/csar.py
```diff
@@ -97,109 +344,35 @@ def package_csar(self, output, service_template=None, csar_format="zip"):
                 meta_file.touch()
                 meta_file.write_text(content)
 
-                return shutil.make_archive(output, csar_format, tempdir)
-        except Exception as e:
-            raise ParseError("Error when creating CSAR: {}".format(e), self)
+                return shutil.make_archive(output_file, csar_format, tempdir)
+        except Exception:
+            import traceback
```
anzoman, line 349: Is this import here because it's only needed for errors? Wouldn't it be nicer to have all imports at the top anyway?

src/opera/parser/tosca/csar.py
```diff
@@ -97,109 +344,35 @@ def package_csar(self, output, service_template=None, csar_format="zip"):
...
-        return self._metadata.get('template_name')
-
-    # prepared for future, currently not used since this is an optional keyname
-    def _get_author(self):
-        return self._metadata.get('template_author')
+        raise OperaError("Cannot unpackage a directory-based CSAR.")
+
+    def members(self) -> List[PurePath]:
+        # queue loop instead of recursion
+        # processed by preorder traversal
```
anzoman, line 357: I'm curious - why preorder traversal? Is this even important?

src/opera/parser/tosca/csar.py
```diff
@@ -1,31 +1,286 @@
...
+
+        if not resolved_path.exists():
+            raise FileNotFoundError("File does not exist at {}".format(path))
+
+        if resolved_path.is_file():
+            return FileCloudServiceArchive(resolved_path)
+        elif resolved_path.is_dir():
+            return DirCloudServiceArchive(resolved_path)
+        else:
+            raise FileNotFoundError("CSARs are either regular files or directories.")
```
anzoman, line 132: Is FileNotFoundError appropriate here. Or is there some better error to describe the wrong CSAR type?

anzoman (changes requested):
@sstanovnik that's great, many thanks for polishing the CSAR class. I have tested some opera commands to make sure that we don't break anything with this PR and everything seems alright. I just have some minor comments and change requests and then that'll be it.

src/opera/parser/tosca/csar.py
sstanovnik: I'll wrap everything at 79, but I'd like to expres interest to change this to something more humane, such as 120.

src/opera/parser/tosca/csar.py
sstanovnik: This is intentional. Previously, every line in the meta file had spaces in front, which worked because YAML ignores consistent leading spaces. However, this isn't a proper TOSCA meta file - per the standard, it should have been parsed as (converting to JSON for clarity:

```json
{
    "                    TOSCA-Meta-File-Version": "1.1Csar-Version: 1.1CreatedBy: xOpera TOSCA orchest...."
}
```

src/opera/parser/tosca/csar.py
sstanovnik: Why? Is this a coding style specific to this project I'm not aware of? I followed the standard stdlib-lib-project structure.

src/opera/parser/tosca/csar.py
sstanovnik: By writing a meta file parser. The format is a bit underspecified though.

src/opera/parser/tosca/csar.py
sstanovnik: Of course. I forgot to move it :)

src/opera/parser/tosca/csar.py
sstanovnik: It's so you know roughly what order the files are in in the result. Reading this comment is simpler than reading the recursion-but-actually-loop below. :)

src/opera/parser/tosca/csar.py
sstanovnik: It doesn't fit, true. I changed it to `OperaError` because it's related to opera, not the filesystem.

src/opera/parser/tosca/csar.py
anzoman: Yes I agree, wrapping lines to 79 chars gives too little space for coding without worrying that you have to do a new line every few minutes or so. I've created a new issue for that: #163.

src/opera/parser/tosca/csar.py
anzoman: Yeah this is the structure that is mostly used within opera right now. I believe that this comes from following [PEP8](https://www.python.org/dev/peps/pep-0008/) guideline:

image

src/opera/parser/tosca/csar.py
anzoman: Aha I see. Yes `TOSCA.meta` is not a YAML. But the code looked really strange as this was moved to the left.

src/opera/parser/tosca/csar.py
anzoman: Oh, very smart move.

src/opera/parser/tosca/csar.py
anzoman: Aha ok, we will se if the meta file will be more spcified in TOSCA Simple Profile in YAML `v2.0`.

src/opera/parser/tosca/csar.py
anzoman: Line 118 in the `csar.py` still exceeds 79. Now that's really annoying. If you want you can fix this limit to 120 before me in a separate PR.

anzoman:
@sstanovnik just a few comments are left. Please take a look at those and then the PR will be ready to merge.

src/opera/parser/tosca/csar.py
sstanovnik: I changed this to concatenated single-quoted strings. It now no longer sticks to the left.

src/opera/parser/tosca/csar.py
sstanovnik: Exactly, the guideline is what I used. If you want, I can change it to your style, but it isn't PEP8.

anzoman: Ok, we can then keep that. All in all, I think that my suggestion for the imports above just looks nicer, but since we are still trying to stick to PEP8 this is ok.

anzoman (approved):
@sstanovnik thanks for your improvements! This is ready to be merged.

PR #154, opened 2020-12-31
Stateless info

sstanovnik:
A WIP because it depends on #153.

Info will now use CSAR metadata to fill out some `info` fields, which will make xopera-saas be able to autofill the service template name even before anything is deployed.

### DIFF
--- a/src/opera/commands/info.py
+++ b/src/opera/commands/info.py
@@ -1,14 +1,16 @@
 import argparse
-import json
 import yaml
 import shtab
 
 from os import path
 from pathlib import Path, PurePath
+from typing import Optional
 
-from opera.error import DataError, ParseError
+from opera.error import DataError, ParseError, OperaError
 from opera.parser import tosca
+from opera.parser.tosca.csar import CloudServiceArchive
 from opera.storage import Storage
+from opera.utils import format_outputs
 
 
 def add_parser(subparsers):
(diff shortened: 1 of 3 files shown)


src/opera/commands/info.py
```diff
@@ -69,6 +76,19 @@ def info(storage: Storage) -> dict:
         status=None
     )
 
+    # stateless autodetect first if possible, which can then be overwritten via state
```
anzoman, line 79: This comment is too long. Please split into two lines.

src/opera/commands/info.py
```diff
@@ -46,7 +53,7 @@ def _parser_callback(args):
 
     storage = Storage.create(args.instance_path)
     try:
-        outs = info(storage)
+        outs = info(args.csar_or_rootdir, storage)
         print(format_outputs(outs, args.format))
```
anzoman, line 57: If you rebase this branch onto master you will see that `format_outputs` are now available as a helper function from `utils.py`. So I propose (as you are already modifying this file) removing `format_outputs` from this file and importing the one from `utils.py` which you can then use.

anzoman (changes requested):
Looks nice @sstanovnik! I propose some minor changes and then we can merge this one.

src/opera/commands/info.py
sstanovnik: done!

sstanovnik:
I've addressed all change requests, but this still depends on #153. I've also finally found the draft button :D

anzoman (dismissed):
@sstanovnik splendid! This will be then merged after #153.

anzoman:
@sstanovnik good news - #153 was merged. Can we remove the draft and merge this one to?

anzoman (approved):
@sstanovnik I approve, thanks again for your work.

Issue #155, opened 2020-12-31
Introduce update command


alexmaslenn:
## Description
In this issue we describe an enhancement for opera to update deployed Topology Template with a new one. Details are described in issue #130
## Current behaviour
Currently, opera does not provide a way to update deployed Topology Template. 
## Expected results
The user is able to run `update` command providing a changed blueprint and inputs in a folder containing existing Service Template that was deployed before. The result of the execution would be undeployment of the nodes that were removed from  Service Template, deployment of the nodes that were added to Service Template and consequential undeployment/deployment of changed nodes.

PR #156, opened 2020-12-31
Add update command

alexmaslenn:
The user is able to run update command providing a changed blueprint and
inputs in a folder containing existing Service Template that was deployed
before. The result of the execution would be undeployment of the nodes
that were removed from Service Template, deployment of the nodes that
were added to Service Template and consequential undeployment/deployment
of changed nodes.

### DIFF
--- a/src/opera/commands/update.py
+++ b/src/opera/commands/update.py
@@ -1,0 +1,133 @@
+import argparse
+import yaml
+import tempfile
+
+from os import path
+
+from opera.error import DataError, ParseError
+from opera.storage import Storage
+from opera.compare.template_comparer import TemplateComparer
+from opera.compare.instance_comparer import InstanceComparer
+from opera.compare.diff import Diff
+from opera.utils import format_outputs, get_template, get_workdir
+from opera.commands.diff import diff_instances
+
+
+def add_parser(subparsers):
+    parser = subparsers.add_parser(
+        "update",
+        help="Update the deployed TOSCA service template and redeploy "
+             "it according to the discovered template diff"
+    )
+    parser.add_argument(
+        "--instance-path", "-p",
+        help=".opera storage folder location"
+    )
+    parser.add_argument(
+        "--inputs", "-i", type=argparse.FileType("r"),
+        help="Optional: YAML or JSON file with inputs "
+             "that will be used for deployment update",
+    )
+    parser.add_argument(
+        "--workers", "-w",
+        help="Maximum number of concurrent update "
+             "threads (positive number, default 1)",
+        type=int, default=1
+    )
+    parser.add_argument(
+        "--verbose", "-v", action='store_true',
+        help="Turns on verbose mode",
+    )
+    parser.add_argument("template",
+                        type=argparse.FileType("r"), nargs='?',
+                        help="TOSCA YAML service template file",
+                        )
+    parser.set_defaults(func=_parser_callback)
+
+
+def _parser_callback(args):
+    if args.instance_path and not path.isdir(args.instance_path):
+        raise argparse.ArgumentTypeError("Directory {} is not a valid path!"
+                                         .format(args.instance_path))
+
+    if args.workers < 1:
+        print("{} is not a positive number!".format(args.workers))
+        return 1
+
+    storage_old = Storage.create(args.instance_path)
+    comparer = TemplateComparer()
+    instance_comparer = InstanceComparer()
+
+    if args.template:
+        service_template_new = args.template.name
+    else:
+        print("Template file for update was not supplied.")
+        return 1
+
+    try:
+        if args.inputs:
+            inputs_new = yaml.safe_load(args.inputs)
+        else:
+            inputs_new = {}
+    except Exception as e:
+        print("Invalid inputs: {}".format(e))
+        return 1
+
+    workdir_old = get_workdir(storage_old)
+
+    try:
+        with tempfile.TemporaryDirectory() as temp_path:
+            storage_new = Storage.create(temp_path)
+            storage_new.write_json(inputs_new, "inputs")
+            storage_new.write(service_template_new, "root_file")
+            workdir_new = get_workdir(storage_new)
+
+            instance_diff = diff_instances(storage_old, workdir_old,
+                                           storage_new, workdir_new,
+                                           comparer,
+                                           instance_comparer, args.verbose)
+
+            update(storage_old, workdir_old,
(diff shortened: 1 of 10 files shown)


alexmaslenn:
Closes #155

src/opera/compare/instance_comparer.py
```diff
@@ -24,6 +24,21 @@ def compare_topology_template(self, topology_template1,
...
 
+    def prepare_update(self, topology_template1, topology_template2,
+                       template_diff):
+        # no need to take relationships into account
+        # as they are always in "initial" state
+        nodes_diff = template_diff.changed.get("nodes", Diff())
+        for node1 in topology_template1.nodes.values():
+            node_name = self._get_template_name(node1)
+            if not nodes_diff.find_key(node_name):
+                node1.set_state("initial", write=False)
```
anzoman, line 35: With these states one can easily make a mistake. I believe that we should create an enum-like python class for storing these states and call them throughout the code. We should create an issue for that. I'm just writing it down here not to forget.

src/opera/instance/base.py
```diff
@@ -57,9 +57,10 @@ def tosca_id(self):
     def state(self):
         return self.attributes["state"].data
 
-    def set_state(self, state):
+    def set_state(self, state, write=True):
         self.set_attribute("state", state)
-        self.write()
+        if write:
+            self.write()
```
anzoman, line 63: For what is this new `write` parameter needed? Just to write the initial state?

src/opera/instance/topology.py
```diff
@@ -53,5 +53,12 @@ def undeploy(self, verbose, workdir, num_workers=None):
     def write(self, data, instance_id):
         self.storage.write_json(data, "instances", instance_id)
 
+    def write_all(self):
+        for node in self.nodes.values():
+            node.write()
```
anzoman, line 58: How does that `write_all` change the current behavior of opera storage?

src/opera/instance/topology.py
```diff
@@ -53,5 +53,12 @@ def undeploy(self, verbose, workdir, num_workers=None):
...
+
     def read(self, instance_id):
         return self.storage.read_json("instances", instance_id)
+
+    def set_storage(self, storage):
+        self.storage = storage
```
anzoman, line 64: Why do you need to (re)set storage?

anzoman (changes requested):
@alexmaslenn nice work! It seems that this is almost done. I've tried the new `opera update` locally and it works and it's really good that we have unit and integration tests also. I just have some minor comments.

src/opera/compare/instance_comparer.py
```diff
@@ -24,6 +24,21 @@ def compare_topology_template(self, topology_template1,
 
         return equal, diff_copy
 
+    def prepare_update(self, topology_template1, topology_template2,
+                       template_diff):
+        # no need to take relationships into account
+        # as they are always in "initial" state
```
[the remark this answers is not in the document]
alexmaslenn: That was exactly my thought as well, but it is a TOSCA standard
https://docs.oasis-open.org/tosca/TOSCA-Simple-Profile-YAML/v1.3/os/TOSCA-Simple-Profile-YAML-v1.3-os.html#_Toc26969445

src/opera/instance/topology.py
alexmaslenn: It does not. It is just a simplification, in case we want to save whole topology state at a time.

src/opera/instance/base.py
alexmaslenn: It is needed if we do not want to save topology state on disk, just keep it in memory.

src/opera/instance/topology.py
alexmaslenn: Basically we have 2 scenarios. 
- we have 2 topologies with different storages, and we want to update from one to another. It is a simple one, at the end we would have all instances as `initial `in the first storage and all `deployed `in the second
- we have 2 topologies but we want to keep only one storage, and we want overwrite first instance states with second states
In case we want to overwrite we assign old storage to a new topology.

src/opera/compare/instance_comparer.py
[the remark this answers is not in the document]
anzoman: Oh, I see ok then we'll leave it as it is

src/opera/instance/topology.py
anzoman: Ok good, I see that you have that `overwrite` parameter - but why don't we expose that as a CLI option `--overwrite/-o`?

src/opera/commands/update.py
```diff
@@ -0,0 +1,132 @@
...
+             "threads (positive number, default 1)",
+        type=int, default=1
+    )
+    parser.add_argument(
+        "--verbose", "-v", action='store_true',
+        help="Turns on verbose mode",
+    )
+    parser.add_argument("template",
+                        type=argparse.FileType("r"), nargs='?',
+                        help="TOSCA YAML service template file",
```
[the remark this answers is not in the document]
alexmaslenn: I do but later in a separate PR. As I can see CSAR logic impl is ongoing and there are a lot of changes incoming
https://github.com/xlab-si/xopera-opera/pull/154/files
we should probably merge them first.

And I'm pretty convinced CSAR support would require additional discussions. Current code state is a little messy in my opinion.
E.g we use both `ZipFile(csar_name, 'r').extractall(csar_dir)` and `shutil.unpack_archive(self._csar_name, output_dir, csar_format)` that basically do the same or not?

src/opera/instance/topology.py
alexmaslenn: Maybe be it is a good idea but again as a separate enhancement. Its better to discuss it first, it may be not as trivial as it seem.

src/opera/commands/update.py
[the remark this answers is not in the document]
anzoman: Yes, I agree.

The functions you mentioned do almost the same - they extract the compressed file, but `unpack_archive` also works with tar compressed CSARs.

anzoman:
@alexmaslenn  I believe that everything is now resolved and fixed. You can now squash the 3 commits into one (do that locally because I think that we don't have that option included here on GH) and then we can finally merge.

alexmaslenn:
@anzoman squash done. Please approve

anzoman (approved):
@alexmaslenn thanks again, I approve. Now, let's go merge this!

PR #159, opened 2021-01-08
Specify README format in setup.cfg

anzoman:
In order to successfully render our README while publishing on PyPI we
need to specify the format of the README as 'text/markdown'. Without
this publishing on PyPI with twine fails as the format defaults to
rst (reStructuredText). And since we use more friendly md (markdown) for
our README we added long_description_content_type metadata to setup.cfg.


cankarm:
Great! I would like to get rid of those failed messages. Let's try how this works.

cankarm (approved):

PR #161, opened 2021-01-13
Major changes in the documentation 

cankarm:
During the development of the xOpera and its tools, a large pile of documentation has been generated. This PR establishes a new table of contents and puts in some stubs that still need to be populated.

The idea is to start with this state and polish it further.


sstanovnik (approved):
The only thing that bothers me here is that CSARs are still packed in the docs. I guess we never got around to fixing that.

Let's just get this out of the way anyway.

cankarm:
@sstanovnik currently we still have CSARs there, later we will put those on TPS and leave only links on docs.

PR #162, opened 2021-01-13
Fix xOpera documentation project name

cankarm:
Fix xOpera docs title and author.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -18,9 +18,9 @@
 
 # -- Project information -----------------------------------------------------
 
-project = u'RADON'
-copyright = u'2020, RADON team'
-author = u'RADON team'
+project = u'xOpera'
+copyright = u'2020, XLAB d.o.o., xOpera team'
+author = u'XLAB d.o.o.'
 
 # The short X.Y version
 version = u''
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -18,9 +18,9 @@
 
 # -- Project information -----------------------------------------------------
 
-project = u'RADON'
-copyright = u'2020, RADON team'
-author = u'RADON team'
+project = u'xOpera'
+copyright = u'2020, XLAB d.o.o., xOpera team'
```
anzoman, line 22: I believe that the year can be updated to 2021 :smiley:

anzoman (commented):
@cankarm looks good, I left just one minor comment.

docs/conf.py
cankarm: Dammit! :D

cankarm: Here you go, @anzoman! ;)

anzoman:
@cankarm I almost forgot -  I have one additional comment - just to be consistent with opera commits can you add a little longer and descriptive commit message (something like Fix documentation title) and then optionally you can add some commit description of you want. And also - we have two commits here now - can you squash them into one?

anzoman (approved):
This can be approved without addressing my comment above because the commits from this PR will be above the ones that are already applying the documentation changes.

Issue #121, opened 2020-10-29
Improve opera documentation and add contribution guide


anzoman:
## Description
The opera's docs are currently available on GitHub pages here: https://xlab-si.github.io/xopera-opera/.

If one looks at the documentation at the moment, he can get info on how to download and run opera TOSCA orchestrator, where opera is/has been used (EU projects), what is the purpose of the tool, past contributors, opera SaaS documentation and so on. This is all very nice but things can get better. I would recommend splitting our documentation into sensible sections where the user would find other information like:

- how every opera command is used exactly and troubleshooting
- which parts of TOSCA does opera support and which are yet to be implemented
- which automation tools are supported and how
- some simple and complex examples based on supported TOSCA entities
- how can users contribute to opera

This last line is very important for the newcomers and users that would want to help and contribute to opera. Since opera's source code can be complex to understand some times, we would need to have some kind of a roadmap that would lead programmers through the different parts of the code. I would also recommend that we modify README and and add a link to the docs there.

I believe that all this can and will make opera a lot more attractive to the new users.

## Steps
We will need to modify rst files that are located in `/docs` folder.

## Current behaviour
Opera has a documentation but it is far from being complete.

## Expected results
To have a more user friendly and informative documentation.

anzoman:
Connected PRs: #137 and #140.

anzoman:
The opera documentation has been updated with a lot of new things here #161. The documentation is available [here](https://xlab-si.github.io/xopera-opera/) and more changes are yet to come. The update plan for opera docs currently includes the following activities on:

- [ ] **add more examples and their description** (AWS, Azure, GCP, OpenFaaS, Docker, Kubernetes etc.), the new TOSCA templates that will be described in the docs could reside in the almost forgotten [`opera-examples`](https://github.com/xlab-si/xopera-examples) GitHub repository which was designed to be the central point for all the opera examples. And because opera might extend in the future, a lot of examples are to come and their maintenance and CI testing would be easier within a separated repo (the connected issue from some time ago is here: https://github.com/xlab-si/xopera-opera/issues/93). So the necessary thing to do would feature moving all (RADON and SODALITE) examples to `opera-examples` repository and then we would need to establish some CI testing for them and later reference and describe them in docs.
- [ ] **describe new opera commands and their usage** targeting `opera update` and `opera diff`.
[...]

PR #165, opened 2021-01-21
Sanity and static analysis checks: PEP8, linters, typing, including code reformatting.

### LINKED ISSUE #163: Increase maximum line length within Python code
## Description
Opera is written in Python and while coding it, we were trying to stick to the [PEP8 coding guideline](https://www.python.org/dev/peps/pep-0008/) that also specifies that lines of code need to be wrapped to 79 characters. This wasn't a problem at first but now we have much more code and functionalities. It is evident that 79 is to little space for coding within opera and the developers need to worry about wrapping long lines. So, I propose extending the line limit from 79 to at least 120 characters which would result in more readable code and easier programming. 

## Steps
We have to reformat all Python files and adjust them to the new limit. This can be done efficiently within some IDE.

## Current behaviour
Lines of Python code are limited to 79 characters.

## Expected results
To extend the line-limit to 120 chars and increase code readability.

____
cc @sstanovnik

sstanovnik:
To help prevent preventable coding errors and to unify code style, we introduce almost all standard linting and static analysis tools applicable to this project:

- flake8
- mypy
- doc8
- pydocstyle
- pylint
- bandit
- shellcheck

This is accompanied by editorconfig, which makes it easier for IDEs to adhere to the styles.

Configurations of different tools are set so that their overlapping checks, such as line lengths, are identical. Static typing is relaxed enough to not require major changes in the codebase (such as requiring all variables to be explicitly typed), as well as other linter configurations, which mostly adhere to the style this project has used before.

CI is updated to perform sanity checks as part of its testing routine. A helper script exists to run all sanity checks at once, as a batch, only exiting with an error code at the end, which simplifies CI execution.

Touches nearly every file in the repository to adhere to sanity checks, most notably:

- line length (80 -> 120)
- other PEP8 checks
- security scanner vulnerability fixes
- static typing
- documentation formatting

This uncovered some bad code that was corrected, such as unreachable branches, plain incorrect logic, use of deprecated and insecure functions etc.

Closes #163.


sstanovnik:
If possible, this needs to be merged before doing anything else, otherwise there will be many ugly merge conflicts. It's a perfect time though, seeing as there are no open pull requests at the moment.

src/opera/template/capability.py
```diff
@@ -1,5 +1,5 @@
 class Capability:
     def __init__(self, name, properties, attributes):
         self.name = name
-        self.properties = properties,
+        self.properties = properties
```
sstanovnik, line 4: Marker: this change is linked to the fix in my other comment.

src/opera/template/node.py
```diff
@@ -109,55 +104,44 @@ def get_property(self, params):
...
+        capabilities = tuple(c for c in self.capabilities if c.name == prop)
         if len(capabilities) > 1:
-            raise DataError("More than one capability is named '{}'.".format(
-                prop,
-            ))
+            raise DataError("More than one capability is named '{}'.".format(prop))
 
         if len(capabilities) == 1 and capabilities[0].properties:
-            return capabilities[0].properties[0].get(rest[0]).data
+            return capabilities[0].properties.get(rest[0]).data
```
sstanovnik, line 126: @anzoman this needed to be fixed because I found a trailing comma (see my other comment) that seemed incorrect.

Please verify that my fix is correct :)

src/opera/template/capability.py
anzoman: That's a good find. The `properties` variable within the `Capability` class is a dict. The unnecessary trailing comma behind  puts the dict into a tuple and this brings only trouble.

src/opera/template/node.py
anzoman: I think that we should be okay here.

anzoman (commented):
@sstanovnik this looks fantastic! Finally we have some checks that can prevent a lot of possible code errors, bugs, vulnerabilities and other stuff. I looked through the changes and everything seems alright and the newly added CI configuration works nicely, so I have just a few minor comments and suggestions.

1. When I open the changed files in the _JetBrains PyCharm_ I see that the IDE is suggesting me some additional changes for the `Pipfile` like this:

image

Should these lines be updated too or is this just something I messed up?

2. I see that `indent_size` for yaml/yml is 2 but if I take a look at `misc/compare-templates/template1.yaml` file I see 4 spaces. Have you forgotten to update some of the files? It's also strange that the testing did not fail.

image

3. My last suggestion would be about commit messages which should be without the ending period to keep the consistency with the previous commits. In the future we will add a contribution guide to our documentation and I will describe this too (linked issue https://github.com/xlab-si/xopera-opera/issues/121).

sstanovnik:
1. I think what the plugin wants is to have dependencies explicity. The current version, I think, loads the new depenencies from either `setup.cfg` or `requirements.txt`. I don't dare touch `Pipfile` in that way, because I have no idea how it works :)
2. Done! I forgot to run code formatting on a couple of folders.
3. I added a sanity check for trailing periods as a separate commit. Never again will you have to remind me of this! Take that! :D

anzoman:
1. Ok. interesting. I think that we should think about  moving from `pipenv` and `Pipfile` to something else. 
2. Great!
3. Magnificent!

anzoman (approved):
@sstanovnik I think everything is okay now and we can merge this one. Thanks for your work again!

src/opera/threading/node_executor.py
```diff
@@ -60,7 +56,7 @@ def process_results(self, results):
             try:
                 future.result()
                 self.processed_nodes.remove(node_id)
-            except Exception as e:
+            except (CancelledError, TimeoutError) as e:
```
alexmaslenn, line 59: Hi @sstanovnik. I was checking the code and stumbled upon this change in exception handling. What was the idea behind it?

sstanovnik: I wanted to avoid having `except Exception` everywhere, so I checked what errors can predictably be thrown. Is this not catching something? If so, this is perfect, since it exposed a code path I didn't know about :)

anzoman: I agree, it's great to have exact types of errors to see what exceptions we can expect in certain part of code, so @alexmaslenn if you found additional exception you can add it to the list.

Overall `opera`'s source code lacks typing in a lot of places and we'll try to gradually add types everywhere to be more developer-friendly.

alexmaslenn: @sstanovnik yes in `future.result()` we finalize execution of asynchronous operation and technically it can raise any exception that underlying function raises. I think is is fixed now.

anzoman: @alexmaslenn ok if this is the case we can then revert back.

alexmaslenn: @anzoman I had already fixed this in the last PR

PR #167, opened 2021-01-23
Make maxsize property optional for TOSCA ObjectStorage normative type

### LINKED ISSUE #166: Possible mistake spotted in ObjectStorage TOSCA normative type definition
## Description
The orchestrator treats TOSCA properties as required by default. This also applies for the properties that are from [TOSCA normative type definitions](https://docs.oasis-open.org/tosca/TOSCA-Simple-Profile-YAML/v1.3/cos01/TOSCA-Simple-Profile-YAML-v1.3-cos01.html#_Toc26969462) (e.g., Node, Relationship, Capability, Artifact, etc.) which are pre-defined and automatically imported ar the beginning of the orchestration. We have recently discovered that `tosca.nodes.Storage.ObjectStorage` normative type might has an error in its definition. The `maxsize` property is described to be optional, so `required` keyname on that property should be explicitly set to false or there should be a default value specified. But none of that is present in the type's definition, which is the following:

```yaml
tosca.nodes.Storage.ObjectStorage:
  derived_from: tosca.nodes.Abstract.Storage
  properties:
    maxsize:
      type: scalar-unit.size
      constraints:
        - greater_or_equal: 0 GB
  capabilities:
    storage_endpoint:
      type: tosca.capabilities.Endpoint
```

This means that opera treats `maxsize` as a required property because we use the normative type definition above.

image

## Steps
We can resolve this by adding `required: false` to `tosca.nodes.Storage.ObjectStorage` normative node type.

## Current behaviour
The `maxsize` property is required on TOSCA `tosca.nodes.Storage.ObjectStorage` normative type but it shouldn't be according to the documentation.

## Expected results
The `maxsize` property on TOSCA `tosca.nodes.Storage.ObjectStorage` should be optional.
_______
### Note
This issue was originally discovered in https://github.com/radon-h2020/xopera-opera/issues/14.

anzoman:
In #166 we found that `tosca.nodes.Storage.ObjectStorage` TOSCA 
normative type has a maxsize property which is described to be optional 
but its type definition doesn't explicitly set required keyname to false, 
so we had to do it in order to prevent that opera would treat it as a 
required TOSCA property which would be against the standard.

Must be merged after #165.
Fixes #166.

### DIFF
--- a/src/opera/stdlib/v_1_3.yaml
+++ b/src/opera/stdlib/v_1_3.yaml
@@ -529,6 +529,8 @@ node_types:
     properties:
       maxsize:
         type: scalar-unit.size
+        # This differs from the definition in section 5.9.10 of the TOSCA standard
+        required: false
         constraints:
           - greater_or_equal: 0 GB
     capabilities:


cankarm (approved):
That's ok, corrected as agreed.

PR #178, opened 2021-02-09
Update tosca.nodes.Database normative type definition

### LINKED ISSUE #177: Fix definition error in `tosca.nodes.Database` normative type
## Description
The orchestrator treats TOSCA properties as required by default. This also applies for the properties that are from [TOSCA normative type definitions](https://docs.oasis-open.org/tosca/TOSCA-Simple-Profile-YAML/v1.3/cos01/TOSCA-Simple-Profile-YAML-v1.3-cos01.html#_Toc26969462) (e.g., Node, Relationship, Capability, Artifact, etc.) which are pre-defined and automatically imported at the beginning of the orchestration. We have recently discovered that `tosca.nodes.Database` (from section 5.9.8) normative type might has an error in its definition. The `port` property is described to be optional, so `required` keyname on that property should be explicitly set to false or there should be a default value specified. 

image

But none of that is present in the type's definition, which is the following:

```yaml
tosca.nodes.Database:
  derived_from: tosca.nodes.Root
  properties:
    name:
      type: string
      description: the logical name of the database
    port:
      type: integer
      description: the port the underlying database service will listen to for data
    user:
      type: string
      description: the optional user account name for DB administration
      required: false
    password:
      type: string
      description: the optional password for the DB user account
      required: false
  requirements:
    - host:
        capability: tosca.capabilities.Compute
        node: tosca.nodes.DBMS
        relationship: tosca.relationships.HostedOn
  capabilities:
    database_endpoint:
      type: tosca.capabilities.Endpoint.Database
```

This means that opera treats `port` as a required property because we use the normative type definition above.

## Steps
We can resolve this by adding `required: false` to `tosca.nodes.Database` normative node type.

## Current behaviour
The `port` property is required on TOSCA `tosca.nodes.Database` normative type but it shouldn't be according to the documentation.

## Expected results
The `port` property on TOSCA `tosca.nodes.Database` should be optional.
_______
### Note
This issue was originally discovered in https://github.com/radon-h2020/xopera-opera/issues/14.

anzoman:
In #177 we found that tosca.nodes.Database TOSCA normative type has
a port property which is described to be optional but its type
definition doesn't explicitly set required keyname to false, so we had
to do it in order to prevent that opera would treat it as a required
TOSCA property which would be against the TOSCA standard.

Fixes #177.

### DIFF
--- a/src/opera/stdlib/v_1_3.yaml
+++ b/src/opera/stdlib/v_1_3.yaml
@@ -494,8 +494,9 @@ node_types:
         description: the logical name of the database
       port:
         type: integer
-        description: >
-          The port the underlying database service will listen to for data.
+        description: the port the underlying database service will listen to for data
+        # This differs from the definition in section 5.9.8 of the TOSCA standard
+        required: false
       user:
         type: string
         description: the optional user account name for DB administration


cankarm (approved):
LGTM.

PR #187, opened 2021-02-19
Preparation for opera notify

anzoman:
This PR is preparing all the things that we need to implement the new `opera notify` CLI command (see issue #108):

- Detect invalid Standard and Configure interface operations (fixes #146)
- Use enum classes to store orchestration constants (resolves #158)
- Prepare to include policies to topology
- Use interface_types too when collecting node's interfaces
- Update policy_triggers example
- Apply policies to nodes when creating topology
- Enable getting properties from policies
- Enable invoking triggers with notify functions
- Supply unit tests for applying polices to nodes

Issue #108, opened 2020-10-08
Allow scaling operations for xOpera


cankarm:
xOpera needs an operation to initiate scaling. In CLI this should be a command ```opera scale -i notification.json``` which would accept a notification message from monitoring as an input. 
In SaaS version of xOpera, the scale would be initiated with a specific API call that would also accept the notification.

This current fix is used to explore options of how scaling should be implemented. The operation will execute a scaling playbook.

anzoman:
Ok, instead of `-i` switch I would propose `-n` in order not to confuse this monitoring notification with inputs. So you specified that this `opera scale` command would execute an Ansible playbook. Which playbook is that? Would user also need to supply path to the playbook (e.g. `opera scale -n notifcation.json scaling-playbook.yaml`). And since this playbook would then execute scaling tasks it would probably need access to the JSON object from `notification.json` to see what monitoring has captured and based on that scaling up/down would be performed. So how would this JSON be passed to the playbook? Just as a normal variable which can be access with Jinja2? @sstanovnik probably you can give some additional info here.

cankarm:
I thought about ```-n``` and for me is the same. I used ```-i``` instead, just because this is an input for scaling. If this content get's into the process as notification or just as a manual execution, I do not mind for now.

anzoman:
I just ran into [TOSCA interface notifications ](https://docs.oasis-open.org/tosca/TOSCA-Simple-Profile-YAML/v1.3/os/TOSCA-Simple-Profile-YAML-v1.3-os.html#DEFN_ELEMENT_NOTIFICATION_DEF) which might be helpful there with scaling since TOSCA notification definitions can be used to transmit external events to the TOSCA orchestrator. There is also an interesting section about [receiving asynchronous notifications](https://docs.oasis-open.org/tosca/TOSCA-Simple-Profile-YAML/v1.3/os/TOSCA-Simple-Profile-YAML-v1.3-os.html#_Toc26969437).

sstanovnik:
Based on what we've discussed this past week, I propose

```shell script
opera notify -n/--notification <notification file, likely but not necessarily json.json>
```

Opera then shouldn't process this but instead pass the notification directly to "the" playbook.

[...]

PR #172, opened 2021-01-29
Documentation updates: CLI and reformating

anzoman:
This PR includes documentation updates that are:

- document `opera diff` and `opera update`
- reformat other documentation content (images, text, code blocks etc.)

This PR is based on #171 and needs to be merged afterwards.

PR #188, opened 2021-02-19
Introduce opera notify

### LINKED ISSUE #32: Add support for policy targets and triggers

anzoman:
With this PR we will supply new opera CLI command called `opera notify`, which will bring the option to inform the orchestrator about the changes after the deployment and to execute various policy trigger actions (Ansible playbooks). This command will also make possible to implement scaling actions with xOpera. (see #108).

This PR has to be merged after #187. 

Closes #32.

### DIFF
--- a/docs/opera_cli.rst
+++ b/docs/opera_cli.rst
@@ -40,32 +40,35 @@ So the simplest way to test ``opera`` is to install it into virtual environment:
 CLI commands reference
 ==========================
 
-``opera`` was  originally meant to be used in a terminal as a client and it
+``opera`` was originally meant to be used in a terminal as a client and it
 currently allows users to execute the following commands:
 
-+---------------------+----------------------------------------------+
-| CLI command         | Purpose and description                      |
-+=====================+==============================================+
-| ``opera deploy``    | deploy TOSCA service template or CSAR        |
-+---------------------+----------------------------------------------+
-| ``opera undeploy``  | undeploy TOSCA service template or CSAR      |
-+---------------------+----------------------------------------------+
-| ``opera validate``  | validate TOSCA service template or CSAR      |
-+---------------------+----------------------------------------------+
-| ``opera outputs``   | retrieve outputs from service template       |
-+---------------------+----------------------------------------------+
-| ``opera info``      | show information about the current project   |
-+---------------------+----------------------------------------------+
-| ``opera package``   | retrieve outputs from service template       |
-+---------------------+----------------------------------------------+
-| ``opera unpackage`` | retrieve outputs from service template       |
-+---------------------+----------------------------------------------+
-| ``opera diff``      | compare service templates and instances      |
-+---------------------+----------------------------------------------+
-| ``opera update``    | update/redeploy template and instances       |
-+---------------------+----------------------------------------------+
-| ``opera init``      | initialize the service template or CSAR      |
-+---------------------+----------------------------------------------+
++---------------------+--------------------------------------------------------+
+| CLI command         | Purpose and description                                |
++=====================+========================================================+
+| ``opera deploy``    | deploy TOSCA service template or CSAR                  |
++---------------------+--------------------------------------------------------+
+| ``opera undeploy``  | undeploy TOSCA service template or CSAR                |
++---------------------+--------------------------------------------------------+
+| ``opera validate``  | validate TOSCA service template or CSAR                |
++---------------------+--------------------------------------------------------+
+| ``opera outputs``   | retrieve outputs from service template                 |
++---------------------+--------------------------------------------------------+
+| ``opera info``      | show information about the current project             |
++---------------------+--------------------------------------------------------+
+| ``opera package``   | retrieve outputs from service template                 |
++---------------------+--------------------------------------------------------+
+| ``opera unpackage`` | retrieve outputs from service template                 |
++---------------------+--------------------------------------------------------+
+| ``opera diff``      | compare service templates and instances                |
++---------------------+--------------------------------------------------------+
+| ``opera update``    | update/redeploy template and instances                 |
++---------------------+--------------------------------------------------------+
+|| ``opera notify``   || notify the orchestrator about changes after the       |
+||                    || deployment and run triggers defined in TOSCA policies |
++---------------------+--------------------------------------------------------+
+| ``opera init``      | initialize the service template or CSAR (*deprecated*) |
++---------------------+--------------------------------------------------------+
 
 The commands can be executed in a random order and the orchestrator will warn and the orchestrator will warn you
 in case if any problems. Each CLI command is described more in detail in the following sections.
@@ -788,7 +791,7 @@ diff
       .. code-block:: console
          :emphasize-lines: 21
 
-         (venv) $ cd tosca/compare-templates
+         (venv) $ cd misc/compare-templates
          (venv) misc/compare-templates$ opera deploy -i inputs1.yaml service1.yaml
          [Worker_0]   Deploying my-workstation_0
          [Worker_0]   Deployment of my-workstation_0 complete
@@ -969,7 +972,7 @@ update
       .. code-block:: console
          :emphasize-lines: 21
 
-         (venv) $ cd tosca/compare-templates
+         (venv) $ cd misc/compare-templates
          (venv) misc/compare-templates$ opera deploy -i inputs1.yaml service1.yaml
          [Worker_0]   Deploying my-workstation_0
          [Worker_0]   Deployment of my-workstation_0 complete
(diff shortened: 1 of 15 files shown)


anzoman:
After #171 and #172 get merged I intend to also add some documentation for `opera notify`.

anzoman:
@sstanovnik #187 has been merged so this PR is now ready for review.

src/opera/commands/notify.py
```diff
@@ -0,0 +1,141 @@
...
+                return 0
+
+    if not args.force and not args.trigger:
+        print("You have not specified which policy trigger to use (with --trigger/-t or --event/-e) "
+              "and in this case all the triggers will be invoked which might not be what you want.")
+        question = prompt_yes_no_question()
+        if not question:
+            return 0
+
+    # read the notification file and the pass it's contents to the library function
```
sstanovnik, line 81:
```suggestion
    # read the notification file and the pass its contents to the library function
```

src/opera/commands/notify.py
```diff
@@ -0,0 +1,141 @@
...
+        "--trigger", "-t",
+        help="TOSCA policy trigger name or event that will invoke all the actions (interface operations) on policy",
+    )
+    parser.add_argument(
+        "--notification", "-n", type=argparse.FileType("r"),
+        help="Notification file (usually JSON) with changes that will be exposed to TOSCA interfaces",
+    ).complete = shtab.FILE
+    parser.add_argument(
+        "--force", "-f", action="store_true",
+        help="Force the action and skip any possible prompts",
```
sstanovnik, line 40:
```suggestion
        help="Skip any prompts and force execution",
```

src/opera/commands/notify.py
```diff
@@ -0,0 +1,141 @@
...
+    )
+    parser.add_argument(
+        "--verbose", "-v", action="store_true",
+        help="Turns on verbose mode",
```
sstanovnik, line 44:
```suggestion
        help="Enable verbose mode",
```

src/opera/commands/notify.py
```diff
@@ -0,0 +1,141 @@
...
+def _parser_callback(args):
+    if args.instance_path and not path.isdir(args.instance_path):
+        raise argparse.ArgumentTypeError("Directory {} is not a valid path!".format(args.instance_path))
+
+    storage = Storage.create(args.instance_path)
+    status = info(None, storage)["status"]
+
+    if not args.force and storage.exists("instances"):
+        if status == "initialized":
+            print("The nodes have not been deployed yet and running notify might have unexpected consequences.")
```
sstanovnik, line 58:
```suggestion
            print("Running notify without previously running deploy might have unexpected consequences.")
```

src/opera/commands/notify.py
```diff
@@ -0,0 +1,141 @@
...
+            question = prompt_yes_no_question()
+            if not question:
+                return 0
+        if status == "interrupted":
+            print("The deployment has been interrupted and running notify might have "
+                  "unexpected consequences on the nodes that have not yet been deployed.")
```
sstanovnik, line 64:
```suggestion
            print("Running notify after an interrupted deployment might have unexpected consequences.")
```

src/opera/commands/notify.py
```diff
@@ -0,0 +1,141 @@
...
+            question = prompt_yes_no_question()
+            if not question:
+                return 0
+        if status == "undeployed":
+            print("All nodes have been undeployed and running notify might have unexpected consequences.")
```
sstanovnik, line 69:
```suggestion
            print("Running notify in an undeployed project might have unexpected consequences.")
```

src/opera/commands/notify.py
```diff
@@ -0,0 +1,141 @@
...
+            question = prompt_yes_no_question()
+            if not question:
+                return 0
+
+    if not args.force and not args.trigger:
+        print("You have not specified which policy trigger to use (with --trigger/-t or --event/-e) "
+              "and in this case all the triggers will be invoked which might not be what you want.")
+        question = prompt_yes_no_question()
+        if not question:
+            return 0
```
sstanovnik, line 79: I would make the trigger parameter mandatory. It's much easier to later make it optional than the other way around.

src/opera/commands/notify.py
```diff
@@ -0,0 +1,141 @@
...
+        return 1
+    except DataError as e:
+        print(str(e))
+        return 1
+
+    return 0
+
+
+def notify(inputs: typing.Optional[dict], storage: Storage, verbose_mode: bool, trigger_name_or_event: str,
+           notification_file_contents: typing.Optional[str]):
```
sstanovnik, line 106: The signature here has `trigger_name_or_event`, but you don't have a `--event` argument. Intentional?

src/opera/commands/notify.py
```diff
@@ -0,0 +1,141 @@
...
+        if trigger_name_or_event:
+            trigger_name_or_event_exists = False
+            for policy in template.policies:
+                for trigger in policy.triggers.values():
+                    if trigger_name_or_event in (trigger.name, trigger.event.data):
+                        trigger_name_or_event_exists = True
+                        break
+
+            if not trigger_name_or_event_exists:
+                raise DataError("The provided trigger name or event: {} does not exists.".format(trigger_name_or_event))
```
sstanovnik, line 136:
```suggestion
                raise DataError("The provided trigger [???or event???] name does not exist: {}".format(trigger_name_or_event))
```

tests/integration/notify/playbooks/configure.yaml
```diff
@@ -0,0 +1,22 @@
...
+      debug:
+        msg: "{{ cpu_lower_bound }}"
+
+    - name: Print out cpu_lower_bound input
+      debug:
+        msg: "{{ cpu_upper_bound }}"
+
+    - name: Print out cpu_lower_bound input
+      debug:
+        msg: "{{ cpu_lower_bound }}"
```
sstanovnik, line 18: Duplicate?

tests/integration/notify/playbooks/configure.yaml
```diff
@@ -0,0 +1,22 @@
...
+  tasks:
+    - name: Start the configuration of monitoring
+      debug:
+        msg: "Configuring monitoring..."
+
+    - name: Print out cpu_lower_bound input
+      debug:
+        msg: "{{ cpu_lower_bound }}"
+
+    - name: Print out cpu_lower_bound input
```
sstanovnik, line 12:
```suggestion
    - name: Print out cpu_upper_bound input
```

sstanovnik (changes requested):

src/opera/commands/notify.py
```diff
@@ -0,0 +1,141 @@
...
+        help="Notify the orchestrator about the changes (after the deployment) and run triggers from TOSCA policies"
+    )
+    parser.add_argument(
+        "--instance-path", "-p",
+        help="Storage folder location (instead of default .opera)"
+    )
+    parser.add_argument(
+        "--inputs", "-i", type=argparse.FileType("r"),
+        help="YAML or JSON file with inputs to override the inputs supplied in deploy",
+    ).complete = shtab.FILE
```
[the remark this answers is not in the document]
anzoman: These are used to override the inputs that were supplied with `opera deploy`. I thought that the option would come in handy when the user wants to change some input values especially when running notify operation. Does this make sense or do you think that inputs can be dropped out of here?

src/opera/commands/notify.py
anzoman: Yes it's intentional to indicate that trigger OR event name should go here. At first I had a separated options for ``--trigger/-t`` and ``--event`` but then I changed my mind because every trigger always has its event in TOSCA. But since event is defined as `The required name of the event that activates the trigger’s action. A deprecated form of this keyname is “event_type”. `, I was not completely sure that only events should be used to filter out triggers, so I added the option to use also the name of the triggers which is probably even better (because YAML will find duplicate trigger names and will prevent that).

tests/integration/notify/playbooks/configure.yaml
anzoman: Fixed.

tests/integration/notify/playbooks/configure.yaml
anzoman: It is! I have removed it.

src/opera/commands/notify.py
anzoman: Reformatted.

src/opera/commands/notify.py
anzoman: Changed as you suggested.

src/opera/commands/notify.py
anzoman: Done.

src/opera/commands/notify.py
anzoman: Refactored.

src/opera/commands/notify.py
anzoman: Fixed.

src/opera/commands/notify.py
```diff
@@ -0,0 +1,141 @@
...
+from opera.error import DataError, ParseError
+from opera.parser import tosca
+from opera.storage import Storage
+from opera.utils import prompt_yes_no_question
+
+
+def add_parser(subparsers):
+    parser = subparsers.add_parser(
+        "notify",
+        help="Notify the orchestrator about the changes (after the deployment) and run triggers from TOSCA policies"
```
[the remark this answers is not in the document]
anzoman: Refactored.

src/opera/commands/notify.py
anzoman: Dunzo.

src/opera/commands/notify.py
anzoman: Updated.

anzoman:
@sstanovnik thanks for the review!. I have resolved most of them and made a few comments back.

anzoman:
@sstanovnik I have also added some docs for `opera notify`, so please review them. I have also made some connected changes in opera-examples repo so I kindly ask you to review https://github.com/xlab-si/xopera-examples/pull/15 PR too.

src/opera/commands/notify.py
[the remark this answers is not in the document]
sstanovnik: I don't think changing inputs via notify makes sense in the context of a monitoring service triggering a scaling operation, because it wouldn't know what to send, and it has no business knowing inputs anyway. We don't really expect to call notify manually, right?

Having inputs here makes things much broader than they need to be IMO. We want `notify` for scaling, and the way we want to do scaling isn't via changing inputs, but by having the implementation figure things out for itself. If we change inputs here, we make "scaling" via inputs the easiest, but incorrect way.

src/opera/commands/notify.py
sstanovnik: Is there a case where someone would want and _need_ to do that? In the context of scaling, I feel some concurrency issues will show up with running multiple scaling operations at the same time. I'd solve that case with something like

```bash
opera notify --trigger "$(opera notify --list)"
```

to trigger many at once, but that's something we'd add later. I suggest keeping things as simple as possible until we know how we actually use `notify`, because I have no clue myself :sweat_smile:

src/opera/commands/notify.py
sstanovnik: If an event and a trigger have the same name, what happens? My concern here is that we're obscuring this already confusing notification/event/trigger terminology even further. If you want to keep this, I'd suggest adding an `--event` alias to the command line argument.

But I'd much rather only have one way of specifying what gets called, because it's simpler to explain. If we keep both, we blur the disinction between event names and trigger names, which makes understanding things harder. That's just my gut feeling though, I don't write TOSCA :)

src/opera/commands/notify.py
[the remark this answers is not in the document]
anzoman: Well, although we are looking more from the xOpera SaaS perspective, calling `opera notify` manually is possible in real use cases. And the user could use it for anything that he wants to on already deployed nodes if the policy on that node allows it. I've prepared some examples in https://github.com/xlab-si/xopera-examples/pull/15.

But I agree that changing inputs in `notify` should not be meant for the purposes of scaling, but rather for something else. And since we don't have any real use case for that I have removed the `--inputs/-i` parameter.

src/opera/commands/notify.py
anzoman: Yes, running multiple triggers that should not be run together on the same node could be a problem (and the user is warned about that). As you put it, invoking all the triggers probably would not be so common, it would be sensible when having different types of nodes (e.g. OpenStack VM, AWS EC2 and so on) and you wanna scale them all at once.

I agree that we don't know yet how `opera notify` will be used (apart from SaaS). So, till then I'm making `--trigger` param mandatory and if we find it problematic we can still make it optional in the future.

src/opera/commands/notify.py
anzoman: A trigger can have the same name as its event and nothing will happen. Each trigger always has an event that activates it. As TOSCA states event represents the required name associated with an interface notification on the identified resource (node).

image

So there's much more meaning in those events and surely we cannot say that trigger is the same as event. But when invoking triggers I don't see the actual reason to distinct these two parameters. So, naturally events invoke triggers and the same would be in SaaS, but as a user I would also want to invoke triggers by their name maybe because some event would activate multiple triggers and I would want just want so I would specify it exactly by its name. Or there could be cases when the user would want to invoke multiple triggers by their name prefix (like all RADON triggers that start with radon.triggers). So I suggest that we keep both options but we focus on the actual trigger events more. If we need to remove trigger name filtering we can do it anytime. I have renamed `--trigger/-t` to `--event/-e`.

docs/opera_cli.rst
```diff
@@ -40,32 +40,35 @@ So the simplest way to test ``opera`` is to install it into virtual environment:
...
+| ``opera package``   | retrieve outputs from service template                 |
++---------------------+--------------------------------------------------------+
+| ``opera unpackage`` | retrieve outputs from service template                 |
++---------------------+--------------------------------------------------------+
+| ``opera diff``      | compare service templates and instances                |
++---------------------+--------------------------------------------------------+
+| ``opera update``    | update/redeploy template and instances                 |
++---------------------+--------------------------------------------------------+
+|| ``opera notify``   || notify the orchestrator about changes after the       |
+||                    || deployment and run triggers defined in TOSCA policies |
```
sstanovnik, line 68: Does this render correctly? The double pipes seem a bit weird.

tests/integration/notify/runme.sh
```diff
@@ -0,0 +1,14 @@
...
+# get opera executable
+opera_executable="$1"
+
+# perform integration test
+$opera_executable validate service.yaml
+$opera_executable deploy service.yaml
+# scale down by calling scale_down_trigger with notification_scale_down.json notification file
+$opera_executable notify -t scale_down_trigger -n files/notification_scale_down.json
+# scale up by calling scale_up_trigger with notification_scale_up.json notification file
+$opera_executable notify -t scale_up_trigger -n files/notification_scale_up.json
```
sstanovnik, line 13: Can you change this to `-e` so we use the alias in a test?

sstanovnik (changes requested):
Just two minor things left.

docs/opera_cli.rst
anzoman: Yes, the second pipe is for the line block. This is a special syntax for [multi-line table rows](https://stackoverflow.com/questions/13504595/multi-line-table-cell-in-restructuredtext). We already us this here: https://xlab-si.github.io/xopera-opera/opera_cli.html#secrets-and-environment-variables and it renders okay.

tests/integration/notify/runme.sh
anzoman: Done

sstanovnik (approved):

sstanovnik:
Thanks!

Issue #173, opened 2021-01-29
BUG: diff_instances does not work if workdir_old != workdir_new


mihaTrajbaric:
Function [opera.commands.diff.diff_instances( storage_old, workdir_old, storage_new, workdir_new,...)](https://github.com/xlab-si/xopera-opera/blob/5aa2c90d24512b2a245954bbd2e2354103a08265/src/opera/commands/diff.py#L157)
should be able to calculate difference between two instance states in two different workdirs. 

This fails because of this two function calls:

```python
# inside diff_instances(...)
 template_old = get_template(storage_old)
 template_new = get_template(storage_new)
```

and implementation of function [opera.utils.get_template(storage)](https://github.com/xlab-si/xopera-opera/blob/5aa2c90d24512b2a245954bbd2e2354103a08265/src/opera/utils.py#L89)
```python
# inside get_template(...)
 ast = tosca.load(Path.cwd(), PurePath(service_template))
```
which works only, if `Path.cwd() == workdir_old == workdir_new`

anzoman:
@mihaTrajbaric thanks for bringing this up! I believe that this issue was known about when we released the first version of opera that included `opera diff` and `opera update`, knowing that we will have to resolve this in the future. Maybe @alexmaslenn can provide more info on how to fix this bug efficiently.

anzoman:
Fixed with #195.

PR #195, opened 2021-03-26
Fix for diff_instances for different workdirs

alexmaslenn:
get_template function in utils.py was not retrieveing template from
working directory, but rather from current directory. This commit fixes
that issue.


alexmaslenn:
Closes #173

anzoman (approved):
Splendid @alexmaslenn! This fix is really welcome as `diff` and `update` will be fully operational within the workdir similar to other CLI commands. Let's merge this one straight away.

PR #197, opened 2021-04-06
xOpera SaaS documentation update: images, Che plugin.

sstanovnik:
I've updated the screenshots to the current version, which looks a bit nicer and includes the credential -> secret rename we did a while back.

Che plugin docs were also missing, so I wrote those up too.

/cc @cankarm


anzoman (approved):
@sstanovnik perfecto! This docs update will surely help the users to imagine how the xOpera SaaS plugin is used within Eclipse Che.

PR #198, opened 2021-04-08
Fix image names for xOpera SaaS docs

anzoman:
In the previous PR (#197), we have updated the docs for our xOpera
SaaS component with numerous from Eclipse Che. We made a small mistake
because images were referenced with the wrong names in the docs and
were therefore not shown properly. This commit will fix this and will
also update the width for images within documentation, because the
images can be a little but smaller for a better view.


sstanovnik (approved):
Thanks for fixing my embarrasing mistake!

PR #200, opened 2021-04-16
Function evaluation for maps and lists

alexmaslenn:
Previously TOSCA functions like get_input or concat that were
used inside maps or lists were not evaluated. This commit solves
this issue.

### DIFF
--- a/src/opera/value.py
+++ b/src/opera/value.py
@@ -48,16 +48,40 @@ class Value:
             raise DataError("Cannot use an unset value: {}".format(key))
 
         if self.is_function:
-            return self.eval_function(instance)
+            return Value.eval_function(self.data, instance)
+
+        if isinstance(self.data, dict):
+            result_map = {}
+            for map_key, value in self.data.items():
+                result_map[map_key] = Value.check_eval_function(value, instance)
+            return result_map
+
+        if isinstance(self.data, list):
+            result_list = []
+            for value in self.data:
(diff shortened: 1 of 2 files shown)


src/opera/value.py
```diff
@@ -48,16 +48,41 @@ def eval(self, instance, key):
...
 
         if self.is_function:
-            return self.eval_function(instance)
+            return Value.eval_function(self.data, instance)
+
+        if isinstance(self.data, dict):
+            result_map = {}
+            for map_key, value in self.data.items():
+                result_map[map_key] = Value.check_eval_function(value,
+                                                                instance)
```
anzoman, line 57: This can be one-liner.

```suggestion
                result_map[map_key] = Value.check_eval_function(value, instance)
```

anzoman (commented):
@alexmaslenn thanks for your contribution, this is update is welcome!

I have left just one comment. After you resolve it you can rebase onto master and squash your three commits into one.

src/opera/value.py
alexmaslenn: @anzoman IDK linter treats it as an error as line length exceeds 79 chars

sstanovnik: We've increased the line length limit to 120, so this isn't an issue any more.

anzoman (approved):
@alexmaslenn thanks for your implementing this useful feature!

PR #201, opened 2021-04-16
Remove xOpera's documentation from this repository

anzoman:
Dear documentation - it's time to pack up the babies and grab the old
ladies since everything goes back to xopera-docs repository. And since
the farewell has to be complete we need to remove all the documentation
files and related stuff. As mentioned the docs will be transferred to
xopera-docs with https://github.com/xlab-si/xopera-docs/pull/3.


cankarm (approved):
Thanks!

Issue #203, opened 2021-04-22
Improve opera info CLI command


anzoman:
## Description
Right now opera info return the following after the successful deployment of the CSAR:

```console
(.venv) $ opera info
content_root: /home/user/Desktop/xopera-examples/.opera/csars/csar
inputs: /home/user/Desktop/xopera-examples/.opera/inputs
service_template: /home/user/Desktop/xopera-examples/.opera/csars/csar/service.yaml
status: deployed
```

For xOpera SaaS and for the general usage inputs could be displayed as real values instead of path to the file.

The next thing is the status, which shows `interrupted` when deploying and undeploying a CSAR, which is not right as it should be deploying/undeploying.  The `interrupted` status is meant to be used when some operation fails or is manually stopped.

## Steps
Update `src/opera/commands/info.py`.

## Current behaviour
Described above.

## Expected results
To improve `opera info` CLI command.

anzoman:
@sstanovnik please express your improvement wishes (for SaaS) here.

sstanovnik:
I think the way forward on this is to have the input values here directly. Here's my though process.

* The user would, I think, primarily need the contents of the inputs to see the parameters this project was deployed with.
* Having the inputs file path specified would have the following two problems
  * The inputs file path is always the same at `.opera/inputs`, making the field a bit useless, especially if we consider the whole `.opera/` folder opaque to the user. All other file paths in the output of `opera info` aren't located in `.opera/`.
  * The user would need to open the file to see the input values, making the process slower.
* In xopera-saas, there would be no secrets directly stored in the inputs, because these are stored as secrets (with a file pointer in the inputs).

I'll get to implementing this and the fix for the `interrupted` status today, expect two PRs :)

sstanovnik:
Closing this due to #204 being merged, which implements the above.

PR #204, opened 2021-05-11
opera info: rendered input values and additional ST/CSAR metadata

sstanovnik:
Related to #203, and also augments the info output with service template and CSAR metadata, which we're already computing.

If the CSAR is invalid, new data is left empty because parting it relies on the CSAR being valid.


anzoman (approved):
Thanks @sstanovnik, looks good and everything seems to work okay when tested locally.

PR #206, opened 2021-05-12
CI for python 3.9 and some sanity and package updates 

### LINKED ISSUE #150: Add support and CI test for Python 3.9
## Description
[Python 3.9.0](https://www.python.org/downloads/release/python-390/) release is available and to keep opera up to date we should check if our tool works with Python 3.9. Additionally we can add a CI workflow that would use Python 3.9 as the executor to test that everything works.

## Steps
Check the opera installation in Python 3.9 virtualenv and add the test for 3.9 to our CI/CD configuration. Also mention that we have the support for 3.9 in `setup.cfg`.

## Current behaviour
Right now opera isn't tested with Python 3.9

## Expected results
To test opera with Python version 3.9 and to establish a new CI job for testing that.

anzoman:
This PR adds or changes the following things:

- CI workflow that would use Python 3.9 as the executor for sanity, unit and integration tests
- Calculate code coverage only when pushing onto master branch
- Update classifiers from setup.cfg 
- Update package versions in Pipfile
- Refactor code to satisfy pylint

Closes #150.

### DIFF
--- a/.sanity-config.ini
+++ b/.sanity-config.ini
@@ -196,7 +196,8 @@ disable=print-statement,
         unused-argument,
         too-few-public-methods,
         too-many-return-statements,
-        too-many-branches
+        too-many-branches,
+        duplicate-code
 
 # Enable the message, report, category or checker with the given id(s). You can
 # either give multiple identifier separated by comma (,) or put this option
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -137,6 +137,9 @@ workflows:
       - test39
       - code_coverage:
           context: opera-code-climate
+          filters:
+            branches:
+              only: master
       - pypi_test_deploy:
           requires:
             - test36
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,6 +83,11 @@ executors:
       - image: circleci/python:3.8.2
     environment:
       PIPENV_VENV_IN_PROJECT: true
+  python39:
+    docker:
+      - image: circleci/python:3.9.4
+    environment:
+      PIPENV_VENV_IN_PROJECT: true
 
 jobs:
   test36:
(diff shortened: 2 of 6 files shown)


.sanity-config.ini
```diff
@@ -196,7 +196,8 @@ disable=print-statement,
         unused-argument,
         too-few-public-methods,
         too-many-return-statements,
-        too-many-branches
+        too-many-branches,
+        duplicate-code
```
sstanovnik, line 200: I don't like disabling these two. They're annoying, but are good indicator that something is unmaintainable. Did a pylint upgrade increase the strictness or is there a specific block that is giving you trouble?

sstanovnik (changes requested):

.sanity-config.ini
anzoman: Yes `pylint` is definitely more strict than a few versions back. And I'm not talking about one duplicated code block but like 50 or more and also imports included (look here for more info: https://app.circleci.com/pipelines/github/xlab-si/xopera-opera/1090/workflows/e1a35c29-d220-403a-a861-319ed75c7a75/jobs/2472) which would mean a lot of effort to reconstruct the code as it also detects duplicates on places that are not duplicates technically. So I tried to mark duplicated blocks with "#pylint: disable: duplicate-code" but it's just not working for me, so I had no other way than disabling these duplicates globally (I have also found this issue: https://github.com/PyCQA/pylint/issues/214 that seems to be connected to our problems).

anzoman: Aha and by the way I didn't disable `too-many-branches` but just `duplicate-code`, maybe a comma after the `duplicate-code` has confused you.

sstanovnik: I've definitely gotten confused by the comma :) I see green, I see new.

Looking at this a bit more, let's keep duplicate-code disabled. It doesn't add anything here because it marks import blocks as duplicates, and that's just insane.

sstanovnik (approved):

PR #209, opened 2021-05-17
Refactor orchestration states for TOSCA nodes

anzoman:
In #203 we found out that orchestration state/status which is returned
by opera info could be improved. This has been done now and xOpera now
provides the following orchestration states (the are also CLI commands
that trigger changing the state in the brackets):

- `initialized` (opera init)
- `deploying` (opera deploy)
- `deployed` (opera deploy)
- `undeploying` (opera undeploy)
- `undeployed` (opera undeploy)
- `error` (opera deploy/undeploy/update/notify)
- `unknown` (for special cases which might happen)

These state are changed during the execution of TOSCA interface
operations (e.g. Ansible playbooks). We had to modify our CLI commands
a little and also our integration tests.

### DIFF
--- a/src/opera/commands/deploy.py
+++ b/src/opera/commands/deploy.py
@@ -59,7 +59,7 @@ def add_parser(subparsers):
     parser.set_defaults(func=_parser_callback)
 
 
-def _parser_callback(args):
+def _parser_callback(args):  # pylint: disable=too-many-statements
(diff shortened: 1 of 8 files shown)


sstanovnik (approved):
This looks good. We'll see if we need any improvements through real usage.

Issue #193, opened 2021-03-24
Opera validate fails to catch missing input


mihaTrajbaric:
Opera validate command with inputs parameter is supposed to validate service template with inputs.

Consider following example:
```yaml
topology_template:
  inputs:
    marker:
      type: string

  node_templates:
    my-workstation:
      type: tosca.nodes.Compute
      attributes:
        private_address: localhost
        public_address: localhost

    hello1:
      type: hello_type
      properties:
        marker: { get_input: marker }
        location: /tmp/playing-opera/hello1
      requirements:
        - host: my-workstation
```

It has input `marker` without default value. If I validate it with input.yaml file without marker field (for instance with empty inputs file), validation would be a success although TOSCA template (in combination with inputs) is not valid:
```shell script
$ opera validate -i inputs.yaml service.yaml
Validating service template...
Done.

```
Of course, it fails during deployment.
```shell script
$ opera deploy service.yaml
[Worker_0]   Deploying my-workstation_0
[Worker_0]   Deployment of my-workstation_0 complete
[Worker_0]   Deploying hello1_0
[Worker_0]     Executing create on hello1_0
Error processing node hello1_0: Cannot use an unset value: marker
```

I think validation should catch errors like this, especially since it works in opposite direction (it throws `Undeclared inputs` for inputs in inputs.yaml file that are not declared in service template)

anzoman:
@mihaTrajbaric good that you brought up this issue. The orchestrator probably doesn't validate any of the TOSCA functions (such as `get_attribute`, `get_property`, or `get_input`) because these are checked and invoked later with `opera deploy`.  But certainly it is necessary to catch errors in those functions. I think that this can be updated along with establishing checks for implementations for TOSCA interface operations - for example for  Ansible playbooks `opera validate` could make use of [check mode and diff mode validating tasks](https://docs.ansible.com/ansible/latest/user_guide/playbooks_checkmode.html) to validate playbooks behind TOSCA, so then the validation would be similar to deploy (just without actually running the exact tasks within the playbooks) and this would also invoke TOSCA functions and the errors would be shown immediately.

PR #210, opened 2021-05-26
Add validation for executors

anzoman:
With these changes we make it possible to validate xOpera's executors
such as Ansible playbooks with `opera validate`. Till now the CLI command
has been offering only the validation of TOSCA templates which was good
but it failed to catch some important errors that might happen during
the deployment (like #193). This new feature now offers the pre-check
for the deployment and undeployment of TOSCA nodes and relationships
since it creates initial instances. We used Ansible check mode to
execute Ansible playbooks in a "dry run" mode.

Must be merged after #209.


src/opera/executor/ansible.py
```diff
@@ -62,6 +63,10 @@ def run(host, primary, dependencies, artifacts, variables, verbose, workdir):  #
             "-e", "@" + vars_file,
             playbook
         ]
+
+        if validate:
+            cmd.append("--check")
```
sstanovnik, line 68: What worries me here is that Ansible can make changes even though it's in check mode, if tasks either

* disable check mode explicitly
* use modules that support check mode, but have it implemented incorrectly

This also isn't a 100% way of checking validity, since tasks that depend on the output of prior tasks are skipped.

Could we get away with using only `--syntax-check`? This would remove the need for the workers argument everywhere, at least.

src/opera/instance/node.py
```diff
@@ -91,6 +92,33 @@ def _configure_requirements(self,
...
+        self.run_operation(OperationHost.HOST, StandardInterfaceOperation.shorthand_name(),
+                           StandardInterfaceOperation.STOP, verbose, workdir)
+        self.run_operation(OperationHost.HOST, StandardInterfaceOperation.shorthand_name(),
+                           StandardInterfaceOperation.DELETE, verbose, workdir)
+        self.reset_attributes()
+        self.write()
+
+        self.validated = True
+        thread_utils.print_thread("  Validation of {} complete".format(self.tosca_id))
+
```
sstanovnik, line 121: It's likely I'm blind, but where is the validation actually enabled in this function?

sstanovnik (changes requested):

src/opera/commands/validate.py
```diff
@@ -11,17 +11,30 @@
...
         help="YAML or JSON file with inputs",
     ).complete = shtab.FILE
+    parser.add_argument(
+        "--workers", "-w", type=int, default=1,
+        help="Maximum number of concurrent deployment threads (positive number, default 1)"
+    )
+    parser.add_argument(
+        "--tosca-only", "-t", action="store_true",
+        help="Validate only TOSCA templates without executors behind them",
+    )
```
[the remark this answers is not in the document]
anzoman: Ok, done.

src/opera/instance/node.py
anzoman: Oh, I forgot to push the changes, now they are present.

src/opera/executor/ansible.py
anzoman: Aha yes Ansible says:

> When ansible-playbook is executed with --check it will not make any changes on remote systems. Instead, any module instrumented to support ‘check mode’ (which contains most of the primary core modules, but it is not required that all modules do this) will report what changes they would have made rather than making them. Other modules that do not support check mode will also take no action, but just will not report what changes they might have made. Check mode is just a simulation, and if you have steps that use conditionals that depend on the results of prior commands, it may be less useful for you. However it is great for one-node-at-time basic configuration management use cases.

But as you mentioned users can disable check mode in tasks, which would mean that opera would still execute them. So, at the end I used only `--syntax-check` without the ability to set workers when validating.

src/opera/instance/node.py
```diff
@@ -80,16 +81,44 @@ def get_host(self):
...
+                                           target_operation, verbose, workdir, validate)
+
+    def validate(self, verbose, workdir):
+        thread_utils.print_thread("  Validating {}".format(self.tosca_id))
+
+        # validate node's deployment
+        self.run_operation(OperationHost.HOST, StandardInterfaceOperation.shorthand_name(),
+                           StandardInterfaceOperation.CREATE, verbose, workdir, True)
+        self._configure_requirements(ConfigureInterfaceOperation.PRE_CONFIGURE_SOURCE,
+                                     ConfigureInterfaceOperation.PRE_CONFIGURE_TARGET, verbose, workdir)
```
sstanovnik, line 103: I think you're missing a `True` as the last parameter for the `validate` argument?

sstanovnik (changes requested):
Just one little thing left.

src/opera/instance/node.py
anzoman: Aha yes, it's true. I've resolved this one.

anzoman:
@sstanovnik thanks for the review, all comments are now resolved.

Issue #221, opened 2021-11-05
Typing errors in service templates contained into examples/ and tests/integration/


philippemerle:
Several typing errors are present in service templates contained into examples/ and tests/integration/.
I already reported and corrected these errors, see the following issues:
  * https://github.com/xlab-si/xopera-examples/issues/28
  * https://github.com/xlab-si/xopera-examples/issues/30
  * https://github.com/xlab-si/xopera-examples/issues/31
  * https://github.com/xlab-si/xopera-examples/issues/33
  * https://github.com/xlab-si/xopera-examples/issues/35
  * https://github.com/xlab-si/xopera-examples/issues/37
  * https://github.com/xlab-si/xopera-examples/issues/38
  * https://github.com/xlab-si/xopera-examples/issues/40
  * https://github.com/xlab-si/xopera-examples/issues/45

I propose to apply all these corrections to files of this repository.

PR #222, opened 2021-11-05
Corrections related to https://github.com/xlab-si/xopera-opera/issues/221

philippemerle:


anzoman (approved):
This is great @philippemerle, big thanks for your contribution! These changes are much appreciated since they will improve our templates and resolve the remaining TOSCA inconsistencies.

PR #226, opened 2021-11-10
Added missed policy targets

philippemerle:
See https://github.com/xlab-si/xopera-examples/issues/44


anzoman (approved):
Looks good, thanks for your contribution @philippemerle.

Issue #230, opened 2021-11-17
Normative networking types missing in src/opera/stdlib/v_1_3.yaml


philippemerle:
Following normative TOSCA networking types are not defined in src/opera/stdlib/v_1_3.yaml:

* tosca.capabilities.network.Linkable
* tosca.nodes.network.Network
* tosca.nodes.network.Port
* tosca.relationships.network.LinksTo
* tosca.relationships.network.BindsTo

My next PR will add them.

anzoman:
Thanks for spotting this @philippemerle!

PR #231, opened 2021-11-17
Corrections for https://github.com/xlab-si/xopera-opera/issues/230

philippemerle:


anzoman (approved):
Looks nice. Thanks @philippemerle for adding these missing TOSCA normative networking types!

Issue #234, opened 2021-12-22
Mixed error output in synchronous failure 


alexmaslenn:
### Current behaviour
In the situation when multiple node operations fail simultaneously, xOpera output becomes mixed and distorted. For instance a multiworker deployment of a service template with 2 nodes that fail at the same moment 
```yaml
tosca_definitions_version: tosca_simple_yaml_1_3

node_types:
  hello_type:
    derived_from: tosca.nodes.SoftwareComponent
    interfaces:
      Standard:
        type: tosca.interfaces.node.lifecycle.Standard
        operations:
          create:
            implementation: playbooks/create.yaml

topology_template:

  node_templates:
    my-workstation:
      type: tosca.nodes.Compute
      attributes:
        private_address: localhost
        public_address: localhost

    hello1:
      type: hello_type
      requirements:
        - host: my-workstation

    hello2:
      type: hello_type
      requirements:
        - host: my-workstation
```
(e.g with a create playbook like this)
```yaml
---
- hosts: all
  gather_facts: false

  tasks:
    - name: Sleep
      command:
        cmd: sleep 5

    - name: Fail
      fail:
        msg: Failed.
```
the output would be
```
{
    {
        "custom_stats": {},
        "custom_stats": {},
        "global_custom_stats": {},
        "plays": [
            "global_custom_stats": {},
            "plays": [
                {
                    {
                        "play": {
                            "play": {
                                "duration": {
                                    "duration": {
                                        "end": "2021-12-22T12:19:31.349852Z",
                                        "end": "2021-12-22T12:19:31.349288Z",
                                        "start": "2021-12-22T12:19:21.156481Z"
                                    },
                                    "start": "2021-12-22T12:19:21.158668Z"                "id": "00155d04-387c-5bcf-eae2-000000000006",
                                },
                                "name": "all"                "id": "00155d04-387c-4e61-53d3-000000000006",
                                "name": "all"
                            },
                        },
```
### Expected behaviour
Error outputs should be provided one after another. A lock object should be introduced to eliminate race condition in thread output.

anzoman:
@alexmaslenn that's a very good find, we need to fix this bug.

Issue #233, opened 2021-12-22
Provide additional info in OperationError


alexmaslenn:
In case of node operation failed `OperationalError` is raised. Currently this exception contains no meaningful info, but it would be convenient to have:
1. Name of the node failed
2. Name of the interface failed
3. Name of the operation failed

This would significantly help to present the users of xOpera as SaaS a nice description of the problem that occurred.  

Additionally it would be great to extract some meaningful info from Ansible output like the name of the failed task, its message and stderr.

anzoman:
Thanks @alexmaslenn, this could be a great enhancement.

PR #235, opened 2022-01-03
Error info extension and output fix

alexmaslenn:


alexmaslenn:
Closes #234 
Closes #233

src/opera/threading/utils.py
```diff
@@ -1,5 +1,18 @@
...
+class SafePrinter:
+    lock = Lock()
+
+    @classmethod
+    def print_lines(cls, wrapper: TextIOWrapper):
+        cls.lock.acquire()
+        print_thread("------------")
+        for line in wrapper:
+            print(line.rstrip())
+        cls.lock.release()
```
anzoman, line 18: Sanity tests fail with `opera/threading/utils.py|14:8|SafePrinter.print_lines |R1732|consider-using-with|Consider using 'with' for resource-allocating operations`, so you could probably change these lines to:

```suggestion
        with cls.lock:
            print_thread("------------")
            for line in wrapper:
                print(line.rstrip())
```

anzoman (changes requested):
Thanks for these fixes @alexmaslenn, it looks okay. I just left one comment regarding the failure of our sanity tests.

src/opera/threading/utils.py
alexmaslenn: Done.

anzoman (approved):
Looks good, @alexmaslenn thanks for your contribution!

PR #241, opened 2022-01-25
Clean up inputs on `--clean-state` deploy

### LINKED ISSUE #240: `opera deploy --clean-state` doesn't reset inputs
When `opera deploy` is run with inputs and then rerun without input, but with `--clean-state` flag, the old inputs are reused.

abitrolly:
Fixes #240

### DIFF
--- a/src/opera/commands/deploy.py
+++ b/src/opera/commands/deploy.py
@@ -183,6 +183,7 @@ def deploy_compressed_csar(
         delete_existing_state: bool
 ):
     if delete_existing_state:
+        storage.remove("inputs")
         storage.remove("instances")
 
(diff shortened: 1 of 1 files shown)


cankarm:
Thanks @abitrolly for this. First we need to finish the *triage* on #240, then we can proceed.

abitrolly:
Yea, I understand. Still learning what xOpera and TOSCA does.

anzoman:
So, as we discussed in #241 this feature will not be incorporated in `opera` orchestrator right now, but we can reopen this pull request anytime in the future if we find out that inputs need resetting while cleaning current orchestration state.

PR #249, opened 2022-02-14
Fix parsing for interfaces and some other minor updates

### LINKED ISSUE #237: Attach playbook to Compute node without defining a new type
Is it possible to execute playbook when `tosca.nodes.Compute` node template is deployed, without defining a new type?

Using modified example from https://docs.oasis-open.org/tosca/TOSCA-Simple-Profile-YAML/v1.3/os/TOSCA-Simple-Profile-YAML-v1.3-os.html#_Toc26969423

```yaml

topology_template:
  node_templates:
    db_server:
      type: tosca.nodes.Compute
      interfaces:
        Standard:
          configure: scripts/my_own_configure.sh
```

`opera validate` errors out with this.

```bash
Validating service template...
mysql.yml:11:11: [InterfaceDefinitionForTemplate] Invalid keys: configure
```

anzoman:
This PR will:

- add fixes for TOSCA interface operations (see #237)
- add more checks for Standard and Configure TOSCA interface operations
- update outputs for execution of primary implementation artifacts
- update file comparison within template comparer
- remove CI/CD run for Python 3.7
- remove the remaining targeted TODOs

Fixes #237.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,11 +68,6 @@ commands:
       - run: git push origin stable
 
 executors:
-  python37:
-    docker:
-      - image: circleci/python:3.7.4
-    environment:
-      PIPENV_VENV_IN_PROJECT: true
   python38:
     docker:
       - image: circleci/python:3.8.2
@@ -90,10 +85,6 @@ executors:
       PIPENV_VENV_IN_PROJECT: true
 
 jobs:
-  test37:
-    executor: python37
-    steps:
-      - run_tests
   test38:
     executor: python38
     steps:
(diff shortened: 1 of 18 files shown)


alexmaslenn (changes requested):
I've checked #237 issue. While validation works with this PR changes, `deploy` operation still does not take into account operation defined in node_template. I believe the problem is here
https://github.com/xlab-si/xopera-opera/blob/49694282db95bb425b988117b90b629f898803ec/src/opera/parser/tosca/v_1_3/collector_mixin.py#L79
`name` should be changed to `op_name` I guess, as `name` variable value is the interface name.

anzoman:
@alexmaslenn thanks I forgot to check that and have now incorporated your suggestion in this PR.

alexmaslenn (approved):

PR #254, opened 2022-04-14
Tosca parser integration

alexmaslenn:
This is the first draft for TOSCA parser integration into the orchestrator. The idea is to get rid of all `template` entities, already defined in TOSCA parser and move all related methods into `instance` classes. This approach is a trade-off between orchestrator concept and code simplicity. E.g. now we would not be able to extract property value from a node template. While technically we should be able to do it without instantiation (property value is already defined after successful parsing), de-facto topology was always instantiated before any data manipulation. `Operation` entity is gone, all execution functionality is moved into `Base` entity. I think we should have an agreement with the concept, before moving further.

I've only yet made changes to `deploy` command, it should work with this example https://github.com/xlab-si/xopera-opera/tree/main/examples/hello `outputs`, `get_attribute`, `concat` and other function do not work yet.

Some additions to TOSCA parser are also required in order this to work, defined here https://github.com/xlab-si/xopera-tosca-parser/tree/tosca-parser-integration. This may be another point of discussion whether we would need some methods defined in Parser, that are not directly used inside Parser.

Another point is to check, whether we should keep one-to-many template instance relationship. Currently it is preserved, but maybe it is time to get rid of it to simplify the code and eliminate confusion.


anzoman:
@alexmaslenn cool, thanks for starting this PR draft, I'll take a deeper look next week. If you can, I would like to ask you to update the code in a way that CI checks will go through.

anzoman (dismissed):
Looks okay, thanks @alexmaslenn for this huge contribution!

